PR #5, opened 2017-10-04
Add continuous integration support

dwhswenson:
Add support for:

- [x] Travis
- [x] AppVeyor
- [x] Coveralls
- [x] MDTraj-dev (check that we still work against MDTraj's newest build)


dwhswenson:
The way that Travis tests the dev version of MDTraj could be sped up (by not installing the stable version then replacing with dev, but instead just installing dev), but the AppVeyor build takes so long that right now, I just don't care.

Anyway, once the AppVeyor builds finish, this is ready to merge.

PR #14, opened 2017-10-09
Docs

dwhswenson:
Adding docs (to put on RTD)


dwhswenson:
still more docs to do, but merging this so RTD can use the current master

PR #2, opened 2017-03-08
Quick fix for the out of index bug

sroet:
This PR fixes the bug mentioned in issue #1. @dwhswenson 

Closes #1.

PR #20, opened 2017-11-01
Add contact_maps example

dwhswenson:
- [x] Update example with Sander's figshare
- [x] Add example to docs
- [x] Test that RTD can work with example in docs
- [x] Check that example is sufficiently informative


dwhswenson:
@sroet : Could you take a look at the example as it is in here and let me know if it seems to explain everything fully enough? Just looking for a second set of eyes to make sure that it is sufficiently useful for someone to use.

https://github.com/dwhswenson/contact_map/blob/b70b029644565a84d569a454395564656b5e3324/examples/contact_map.ipynb

sroet:
Looks pretty good, the only thing missing in my opinion is a clear pointer to the `.df` attribute (used in cell 32). The `.DataFrame` type is mentioned under the figure, but I read over the '.df' the first time I read the code and no clear connection between `DataFrame` and `.df` is made.

Also, I would print all 3 assigned variables in cell 13, especially since GTP201 is selected as residue 166 (not a simple 0,1 count difference and the reason is explained a lot later)  

I also had a couple style remarks/questions:
> "Otherwise the trajectory would swap the single frame, and there would be no blue in that picture!"

I don't understand what this swap is, did you mean swamp?

> "First we look at the frames that are much more important in the trajectory than the frame. Then we look at the frames that are more important in the frame than the trajectory." 

Shouldn't this be mentioning 'contacts' instead of  'frames'?

> "The query atoms are the atoms which appear on the y-axis of the plot. The haystack atoms are those along the x-axis of the plot." 

is this true? the plots look symmetric to me .

>"Note: for non-protein contact"

shouldn't this be 'contacts'?

That was all that I noticed, I hope it is helpful.

dwhswenson:
@sroet: Thanks! Good catches on all of them.

>> "The query atoms are the atoms which appear on the y-axis of the plot. The haystack atoms are those along the x-axis of the plot."
>
> is this true? the plots look symmetric to me .

This one is your fault! (#2)

PR #28, opened 2017-11-05
Concurrence

dwhswenson:
When looking at the behavior of contacts, it is often not only important to see how often a contact occurs, but also what other contacts occur simultaneously. These simultaneous contacts (which we refer to as "concurrence") can be used to define stable states.


contact_map/concurrence.py
```diff
@@ -0,0 +1,246 @@
...
+    Contact concurrences measure what contacts occur simultaneously in a
+    trajectory. When defining states, one usually wants to characterize
+    based on multiple contacts that are made simultaneously; contact
+    concurrences makes it easier to identify those.
+
+    Parameters
+    ----------
+        values : list of list of bool
+            the whether a contact is present for each contact pair at each
+            point in time; outer list is length number of frames, inner list
```
sroet: this is the other way around outer list is length contact pair and inner list is frame

contact_map/concurrence.py
```diff
@@ -0,0 +1,246 @@
...
+        cutoff, in nm. Should be the same as used in the contact map.
+    """
+    def __init__(self, trajectory, atom_contacts, cutoff=0.45):
+        # TODO: the use of atom_contacts as input from most_common is weird
+        atom_pairs = [[contact[0][0].index, contact[0][1].index]
+                      for contact in atom_contacts]
+        labels = [str(contact[0]) for contact in atom_contacts]
+        distances = md.compute_distances(trajectory, atom_pairs=atom_pairs)
+        vector_f = np.vectorize(lambda d: d < cutoff)
+        values = list(map(list, zip(*vector_f(distances))))
```
sroet: please add a comment for this magic

contact_map/concurrence.py
```diff
@@ -0,0 +1,246 @@
...
+            "resid " + str(idx) + " " + select
+        )
+        for res_A, res_B in residue_pairs:
+            atoms_A = select_residue(res_A.index)
+            atoms_B = select_residue(res_B.index)
+            atom_pairs = itertools.product(atoms_A, atoms_B)
+            distances = md.compute_distances(trajectory,
+                                             atom_pairs=atom_pairs)
+            min_dists = [min(dists) for dists in distances]
+            values.append(list(map(lambda d: d < cutoff, min_dists)))
```
sroet: list comprehension should be more readable

contact_map/concurrence.py
sroet: and this could be changed into: `list(map(list, vector_f(distances).T)` 
as you do want to map the transpose of the array to a list and I suspect this is more performant as it used numpy instead of zip. But I didn't check my suspicion

contact_map/concurrence.py
```diff
@@ -63,6 +65,41 @@ def __getitem__(self, label):
...
+    atom_or_res : string
+        whether to treat this as an atom-based or residue-based contact;
+        allowed values are "atom", "res", and "residue"
+
+    Returns
+    -------
+    list :
+        list in the format of ``ContactCount.most_common()``
+    """
+    if isinstance(contact_input, contact_map.ContactFrequency):
```
sroet: shouldn't this be `isinstance(contact_input, contact_map.ContactObject)` ? As this statement would be `False` for a `ContactMap` or `ContactDifference`

contact_map/concurrence.py
```diff
@@ -0,0 +1,281 @@
...
+        allowed values are "atom", "res", and "residue"
+
+    Returns
+    -------
+    list :
+        list in the format of ``ContactCount.most_common()``
+    """
+    if isinstance(contact_input, ContactObject):
+        contact_input = contact_input.contacts[atom_or_res]
+    if isinstance(contact_input, contact_map.ContactCount):
```
sroet: can you put an extra blank line, this tripped me up and I read `if`, `elif`

PR #36, opened 2018-05-01
Drop Python 3.4 support [travis]

dwhswenson:
`pandas` no longer has wheels for Py-3.4, so we can't pip-install. We require pandas for serialization/deserialization, which is an essential (not optional) feature. Therefore we can no longer support 3.4. Also, MDTraj doesn't test against 3.4 anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: python
 python:
   - '3.6'
   - '2.7'
-  - '3.4'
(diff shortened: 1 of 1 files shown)


dwhswenson:
Note that we should still work with older versions of `pandas` that *do* support Py-3.4. The problem is that we don't want to pin to old versions just so that our tests pass.

PR #40, opened 2018-07-10
fix dask travis error

sroet:
this will try to fix the timeout error on travis, first try is to set the timeout in the client


sroet:
from the [Travis common build problems](https://docs.travis-ci.com/user/common-build-problems/):  "More than two to four processes should be fine, beyond that, resources are likely to be exhausted." So implementing the 4 workers cluster should be a reliable way of fixing this error

dwhswenson:
Passed 4/4 retries ... will merge it and see if it is a fix

PR #41, opened 2018-07-31
Add Python 3.7 support

dwhswenson:
No guarantee this will work; depends on whether MDTraj properly supports 3.7.

~Leaving 3.6 as the "canonical" version for now; that will probably change once we're sure that this works, though.~ "Canonical" version is 3.7 (although I'm not sure if we're using that for anything now; for OPS, that's where we build/deploy docs from.)

Also, update copyright date on docs.


dwhswenson:
Looks like this has to wait on miniconda to update to include 3.7 (or it would require significant rewriting of the .travis.yml). I assume this will happen with the next release of miniconda. They seem to release every few months or so on average; last release was June 6 (https://repo.continuum.io/miniconda/).

dwhswenson:
Current AppVeyor error is, I believe, because MDTraj hasn't released for 3.7. However, I don't understand why I'm not getting a Python 3.7 test in Travis with this.

dwhswenson:
Travis passes, and the 3.7 builds are actually running 3.7! (Now that there's a 3.8 beta release, I guess.)

AppVeyor is still having trouble. I'm getting close, but think it will require pinning versions of conda and/or conda-build.

dwhswenson:
It works! This is ready for review (finally!)

appveyor.yml
```diff
@@ -21,24 +21,26 @@ environment:
...
 
 test_script:
-  - "%CMD_IN_ENV% conda build --quiet ci\\conda-recipe"
+  #- "%CMD_IN_ENV% activate base"
+  #- "%CMD_IN_ENV% python --version"
+  #- "%CMD_IN_ENV% python -c \"import struct; print(struct.calcsize('P') * 8)\""
+  #- "%CMD_IN_ENV% conda build --quiet ci\\conda-recipe"
+  - activate base
+  - python --version
+  - python -c "import struct; print(struct.calcsize('P') * 8)"
```
sroet: Is this leftover debug code or actually usefu/required?

dwhswenson: This verifies that we're on the right architecture (32-bit vs 64-bit), and so plays the same sort of role as the `python --version` above it. It dates back to some very old MDTraj version of the AppVeyor script. That said, I don't know how much longer we'll support 32-bit Windows. MDTraj stopped releasing for 32-bit a few releases ago, so I think we'll do the same soon.

PR #43, opened 2018-11-10
Easy access to haystack/query residue indices and range

dwhswenson:
The example notebook previously showed how to find the residues for query/haystack atoms, and then how to get the max/min needed for plotting that subset. But that was more complicated than it needed to be, especially because I'm finding that those "zoomed in" contact maps are something that I want frequently.

This PR adds:

* `query_residues` / `haystack_residues` : list of the residue indices for the query/haystack
* `query_residue_range` / `haystack_residue_range` : max and min index values of the above

This makes it easy to adjust the plot axes with, e.g.,

```python
ax.set_xlim(*contacts.haystack_residue_range)
```

Includes tests; updates example notebook.


examples/contact_map.ipynb
```diff
@@ -591,7 +591,9 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "### Using a different query"
+    "### Using a different query\n",
+    "\n",
+    "MDTraj has allows queries based on different numbering systems: `resid` and `resSeq`. The `resid` is the internally-used residue number: it starts from 0 and is always continuous. On the other hand, `resSeq` is the residue number given in the PDB, which usually starts from 1 (and is the number we usually refer to in literature). More details can be found in the [documentation on MDTraj's atom selection language](http://mdtraj.org/latest/atom_selection.html)."
```
sroet: either 'has' or 'allows' not 'has allows'

dwhswenson:
Made some implementation changes: I decided that these are better as properties, and that `haystack_residues`/`query_residues` should return the residues themselves (the `mdtraj.core.topology` objects) instead of just the indices.

@sroet: I think you have merge access now -- feel free to merge this after appveyor passes!

PR #44, opened 2018-11-15
Handle dask TimeoutError exception in tests

dwhswenson:
We're still occasionally getting an error when dask fails to start up a cluster (not as often as before #40, but still happens sometimes).

This PR changes the `distributed.LocalCluster` startup so that we can retry a fixed number of times. For now, I'm doing 3 retries. In addition, if all the retries fail, this gives a skip instead of a fail.

If all the retries fail in both the Travis jobs where we try dask, then we'll see a drop in coverage, but I think this will also mean that our coverage will be more stable.

Importantly, the skip only happens if the setup fails. And if the setup is failing consistently (e.g., due to an [unlikely!] API change in dask), I hope we'll notice the drop in coverage.

### DIFF
--- a/contact_map/tests/test_dask_runner.py
+++ b/contact_map/tests/test_dask_runner.py
@@ -6,6 +6,22 @@
 from .utils import *
 from contact_map.dask_runner import *
 
+def dask_setup_test_cluster(distributed, n_workers=4, n_attempts=3):
+    """Set up a test cluster using dask.distributed. Try up to n_attempts
+    times, and skip the test if all attempts fail.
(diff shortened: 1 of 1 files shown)


dwhswenson:
Codeclimate found the complexity at 6 to be too much. I approved it anyway -- (1) The complexity comes from catching multiple error conditions within a trivial loop, so I don't think the loop multiplier should count; (2) It's in tests anyway.

(Honestly, I think the hard limit should be more like 10, but trying to keep the complexity down is good.)

PR #45, opened 2019-01-15
Fix problems with pytest

dwhswenson:
Travis is currently installing old versions of pytest with a newer version of pytest-cov, causing errors in tests. This PR is intended to fix that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,6 +36,7 @@ before_install:
 install:
   - source ci/pip-install/install_requirements.sh
   - pip install -e .
+  - pip search pytest  # TODO TEMP
   - pip install -r ci/pip-install/testing_requirements.txt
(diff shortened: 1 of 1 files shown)


dwhswenson:
Apparently the problem was that Travis already had pytest (3.3) installed. Recent versions of pytest-cov require pytest>3.6. Our request to install pytest as part of our testing requirements was met with "already installed" (at 3.3). The solution is to request up-to-date versions of all testing packages (add a `-U` to the pip command).

Better dependency resolution would help (e.g., using conda). But I want to use pip installs for the testing environment on Travis (we use conda on AppVeyor).

I assume this will pass the AppVeyor builds, since this problem doesn't exist there. Once it does, I'll merge. @sroet, you're welcome to take a look, but I don't think the 2-character change requires a review!

PR #46, opened 2019-01-15
Release 0.3.2

dwhswenson:
Quick fix for problems in our testing procedure. See #45 for details.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ before_install:
 install:
   - source ci/pip-install/install_requirements.sh
   - pip install -e .
-  - pip install -r ci/pip-install/testing_requirements.txt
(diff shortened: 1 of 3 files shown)


dwhswenson:
Had to bump version to 0.3.3. The autorelease process is extremely conservative in its testing. It uploads to the test.pypi.org server, then re-downloads and runs tests. Unfortunately, if there's a problem in the test procedure (as there was here -- the same problem that this release will fix) then the test will fail. And pypi doesn't let you replace a package using the same version number.

Honestly, I could merge this while failing, and it would be fine. Or another approach is to do a post-release, i.e., 0.3.2.post1. But I'm fine just skipping the 0.3.2 and going straight to 0.3.3.

sroet:
LGTM feel free to merge (I agree on the skip part)

PR #47, opened 2019-01-29
Add example for concurrences

dwhswenson:
This provides an example using the contact concurrences introduced in #28. Although I prefer to get example trajectories from figshare, this specific case is a little too vaguely sourced (I got the force field from someone and have an incomplete description of how it was parametrized) to put up on figshare until there's a paper out on it. However, the purpose here is not the specific science behind the trajectory, but to use that trajectory as an illustration of contact concurrences. In the future we might replace this with something to be hosted on figshare.

This also does a little bit of cleanup of the concurrence plot output (removes misleading yticks).


dwhswenson:
This also adds the ability to customize the concurrence plotter output with matplotlib keywords (e.g., to increase the `markersize`) and it updates a few things in the docs (installing optional packages, and a better intro for the examples sections, now that there is more than 1 example).

dwhswenson:
I'm happy with this now -- merging!

Issue #48, opened 2019-03-15
Always show at least one pixel for plotting


dwhswenson:
It is currently possible (especially for atom-atom plots) for the size of one contact to be smaller than a single pixel. The result is that many contacts aren't actually shown. This is very misleading.

We should update the contact object plot such that the dimensions of the patch is guaranteed to be at least 1 pixel. Currently it gives 1 unit on the axis; we need to figure out how many units on the axis correspond to 1 pixel (via matplotlib's resolution).

I think it is fine if we slightly overestimate; this will be a min function, so it will still give a distance of 1 pixel if `pixel_per_axis_unit` > 1.

sroet:
> I think it is fine if we slightly overestimate; this will be a min function, so it will still give a distance of 1 axis unit if axis_per_pixel > 1.

shouldn't this be "still give a distance of 1 pixel"?

dwhswenson:
> shouldn't this be "still give a distance of 1 pixel"?

Corrected in edited text.

sroet:
> `if pixel_per_axis_unit` > 1

this should be smaller than (<) 1, right?

sroet:
I have been looking into this, but I can not reproduce the problem when saving to a PDF

In the branch: https://github.com/sroet/contact_map/tree/always_1_pixel, I have added the `**kwargs` argument to `plot()` which passes through to matplotlib. 

there is no difference between the pdf saved from a `dpi=200` and a `dpi=50` plot.
[atoms.pdf](https://github.com/dwhswenson/contact_map/files/3025503/atoms.pdf)
[atoms2.pdf](https://github.com/dwhswenson/contact_map/files/3025504/atoms2.pdf)

Now, there is a difference between the plots if they are saved to `png`:
![atoms.png](https://user-images.githubusercontent.com/19587267/55277068-32430500-52fc-11e9-86dc-082a4bbb094b.png)
![atoms2.png](https://user-images.githubusercontent.com/19587267/55277071-33743200-52fc-11e9-92d9-8b06681350d2.png)

but this is fixed if  `dpi=200` argument is passed to `plt.savefig()`
![atoms3](https://user-images.githubusercontent.com/19587267/55277186-84d0f100-52fd-11e9-8cbc-2f1d1e1690d3.png)


So this looks a lot like this is a compression during the rendering from matplotlib, however I will try to see if I can fix that.

dwhswenson:
[...]

PR #50, opened 2019-03-30
Warn when there is less than 1 pixel per contact

sroet:
This PR allows for `**kwargs` to be passed down from `ContactCount.plot()` to `plt.subplots()` giving the user more power over the plotting

~~Also, this makes sure that at least 1 pixel is drawn per contact, even if it overlaps with other contacts Potentially solving #48~~

~~I am not to happy about the current effect that if the initial plot is made with a to low dpi, all the figures saved from that plot also look bad. This might not be the implementation we want~~

This raises a `RuntimeWarning` if the figure is to small to show all the contacts

- [API BREAK] This also combines the `vmin` and `vmax` keyword arguments of the `ContactCount.plot()` function into `vrange=(-1.0, 1.0)` in order to obey codeclimate complexity


sroet:
This will implement the first option of: https://github.com/dwhswenson/contact_map/issues/48#issuecomment-478257882.

sroet:
@dwhswenson This is ready for review, 
I updated the notebook to use the correct `vrange` instead of `vmin/vmax`
Also added a cell with possible fixes to the warning:
- let jupyter output a pdf
- save as a pdf
- increase figsize
- increase dpi

sroet:
Apparently the `kwargs` are not past through is `MPLBACKEND=PS`, let's see if `SVG` works on travis (it works locally)

contact_map/contact_count.py
```diff
@@ -98,19 +100,51 @@ def df(self):
...
+            msg = ("The number of pixels in the figure is insufficient to show"
+                   " all the contacts.\n Please save this as a vector image "
+                   "(such as a PDF) to view the correct result.\n Another "
+                   "option is to increase the 'dpi' (currently: "+str(dpi)+"),"
+                   " or the 'figsize' (curently: "+str((figwidth, figheight)) +
+                   ").\n Adviced minimum amount of pixels = "
+                   + str((self.n_x, self.n_y))+" (width, height).")
+            warnings.warn(msg, RuntimeWarning)
+
+    def plot(self, cmap='seismic', vrange=(-1.0, 1.0), with_colorbar=True,
```
dwhswenson: Go ahead and keep `vmin`/`vmax`. We'll tell codeclimate that this is an exception (in general, when most arguments are minor optional tweaks of behavior, I don't worry about the too-many-arguments complaint -- so in my mind, `with_colorbar` and `kwargs` don't really count). The reason to keep them is that `vmin`/`vmax` are the standard names in matplotlib, so we want to keep closer to standard matplotlib terminology.

contact_map/contact_count.py
```diff
@@ -98,19 +100,51 @@ def df(self):
...
-        vmin : float
-            minimum value for color map interpolation; default -1.0
-        vmax : float
-            maximum value for color map interpolation; default 1.0
-
+        vrange : tuple
+            (minimum, maximum) value for color map interpolation;
+            default (-1.0, 1.0)
+        **kwargs
+            All aditional keyword arguments to be passed to the
```
dwhswenson: typo: `additional` (missing d)

contact_map/contact_count.py
```diff
@@ -98,19 +100,51 @@ def df(self):
...
+        ypixels = dpi*figheight
+
+        # Check if every value has a pixel
+        if xpixels/self.n_x < 1 or ypixels/self.n_y < 1:
+            msg = ("The number of pixels in the figure is insufficient to show"
+                   " all the contacts.\n Please save this as a vector image "
+                   "(such as a PDF) to view the correct result.\n Another "
+                   "option is to increase the 'dpi' (currently: "+str(dpi)+"),"
+                   " or the 'figsize' (curently: "+str((figwidth, figheight)) +
+                   ").\n Adviced minimum amount of pixels = "
```
dwhswenson: "Recommended minimum" (better word choice, and typo in current spelling anyway)

contact_map/contact_count.py
```diff
@@ -98,19 +100,51 @@ def df(self):
...
+        xpixels = dpi*figwidth
+        ypixels = dpi*figheight
+
+        # Check if every value has a pixel
+        if xpixels/self.n_x < 1 or ypixels/self.n_y < 1:
+            msg = ("The number of pixels in the figure is insufficient to show"
+                   " all the contacts.\n Please save this as a vector image "
+                   "(such as a PDF) to view the correct result.\n Another "
+                   "option is to increase the 'dpi' (currently: "+str(dpi)+"),"
+                   " or the 'figsize' (curently: "+str((figwidth, figheight)) +
```
dwhswenson: typo: should be "currently"

contact_map/contact_count.py
sroet: like this, or without the `amount of pixels`?

sroet:
> Sorry for long delay in reviewing; I was running the ESDW this week.

No problem, ran through your comments, please have another look.

contact_map/contact_count.py
dwhswenson: with `amount of pixels`; I just meant that "Adviced" is wrong

dwhswenson:
All looks good! Will merge once AppVeyor officially approves. (And for some reason codeclimate didn't even complain?)

sroet:
> All looks good! Will merge once AppVeyor officially approves. (And for some reason codeclimate didn't even complain?)

I might have misinterpreted a `code complexity` complaint as `to many attributes`

dwhswenson:
@sroet : In the end, this was not an API break, right? I think you undid the original API-breaking behavior. (Prepping a release right now.)

sroet:
> @sroet : In the end, this was not an API break, right? I think you undid the original API-breaking behavior. (Prepping a release right now.)

Yes, this was only added functionality (pass through of `**kwargs` in the `plot` function to `matplotlib.subplots`) and bugfixes (prevents incorrect representations).

PR #54, opened 2019-05-22
Fix dask+atom_slice

sroet:
Due to an discrepancy between `DaskFrequency` and it's parent class (`ContactMap`) `__init__()`  , we tried to convert indices back to unsliced two times:
*  `DaskFrequency` calls the `super.__init__`:
*  `ContactFrequency.__init__()` converts to sliced indices, calls `_build_contact_map` and converts back to real indices

**However** `DaskFrequency._build_contact_map` constructs new `ContactFrequency` objects and sums them (in unsliced indices), so after this the `ContactFrequency.__init__()` tries to convert these already correct indices back again.

In order to solve this I moved the conversion from sliced to regular indices as part of the `_build_contact_map` routine of `ContactFrequency`.

fixes #51

PR #35, opened 2018-04-25
Use atom slice

sroet:
This uses atom_slice to cut down a trajectory before computing the neighborlist. This should speed up contactmap/frequency calculations for real systems
The code is not yet completely implemented, but is only tested at one point, and not yet benchmarked for now.

TODO:
 - [x] write documentation
 - [x] write proper tests
 - [x] update example
 - [x] do benchmarking

EDIT: found the first bug already (only converting the list/ not the numpy arrays)

PR #55, opened 2019-06-03
Release 0.4.0

dwhswenson:
The two main enhancements in `contact_map` 0.4.0 are:

1. "Atom slicing": By only processing the atoms that are used, the contact map calculation is now much faster in most common cases (e.g., allowing you to ignore all waters for a protein contact map).

2. Contact concurrences: This is a new way of visualizing contacts during a trajectory, which assists in identifying specific sets of contacts that occur simultaneously. That can be useful for identifying stable states in molecular systems.

# New features
* Easy access to haystack/query residue indices and range (#43)
* Use atom slice (#35) @sroet #performance
* Contact concurrences (#28)

# Bugs fixed
* Fix dask+atom_slice (#54) @sroet

# Miscellaneous improvements
* Improvements to documentation and examples (#47, #52)
* Warn when there is less than 1 pixel per contact (#50) @sroet
* Improvements on dask testing procedures (#40 [@sroet], #44)
* fixed pandas future warn about as_matrix (#37) @sroet


dwhswenson:
We've been way overdue for a release; features like the contact concurrences that are shown in the docs aren't available through pip/conda-installed versions!

I'm holding off on merging this until @sroet has a chance to tell me whether #50 actually was an API break or not (if so, we need to include that in the release notes). Merging should auto-update pypi, and then I can use that to update conda-forge.

sroet:
LGTM, feel free to merge after altering/removing the API break ;)

PR #56, opened 2019-06-04
Bump to 0.5.0.dev0

dwhswenson:
This will complete the release of 0.4.0; now we're working on 0.5.0!


.travis.yml
```diff
@@ -44,7 +44,7 @@ install:
 script:
   - export MPLBACKEND=SVG
   - python -c "import contact_map"
-  - python autorelease_check.py --branch ${TRAVIS_BRANCH} --event ${TRAVIS_EVENT_TYPE}
+  - python autorelease_check.py --branch ${TRAVIS_BRANCH} --event ${TRAVIS_EVENT_TYPE} --allow-patch-skip  #TODO remove allow-patch-skip
```
sroet: What does this alteration do? (I assume this is an autorelease issue?)

dwhswenson: Actually I'm just about to push a commit commenting that out (and giving a slightly more detailed comment as to what it does). Long version is this: By default, `autorelease` does not allow you to skip an expected version, i.e., you can't go from 1.1 to 3.0 (skipping 2.x) or from 0.4.0 to 0.6.0. This includes patch releases: you can't go from 0.3.1 to 0.3.3.

However if there's a problem when deploying to testpypi, you *must* skip a version. This is because you can't modify a package that has been published on testpypi. We did this: there was a bad version released as 0.3.2 on testpypi, and but GitHub releases, pypi, and conda-forge never saw 0.3.2, only 0.3.3.(`autorelease` checks the GitHub releases.)

Anyway, this flag disables the error that autorelease would raise for skipping a patch-level version, so that we can do exactly this thing. I'm commenting it back out because we should, in general, not skip versions.

sroet: Thanks for the explanation.

PR #63, opened 2019-10-20
Fix coveralls reporting

dwhswenson:
Coveralls is reporting each build separately, instead of combining them. There are multiple instances of each file when looking at coverage data. This is trying to fix that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,9 @@ python:
   - '3.6'
   - '3.7'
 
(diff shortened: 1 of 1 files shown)


dwhswenson:
Apparently I completely forgot that I hadn't merged this PR. It seems like it might have fixed the problem, so I'll go ahead and merge, which should mean that other PRs can actually get the full coverage they deserve!

PR #64, opened 2019-11-11
Fix for lack of unitcell; better colorbar scaling

nffaruk:
The colorbar parameters are taken from [here](https://stackoverflow.com/a/26720422) to provide better scaling to square aspect ratio plots (which I also think should be the default aspect ratio)

### DIFF
--- a/contact_map/contact_map.py
+++ b/contact_map/contact_map.py
@@ -1,6 +1,8 @@
 """
 Contact map analysis.
 """
+# Maintainer: David W.H. Swenson (dwhs@hyperblazer.net)
+# Licensed under LGPL, version 2.1 or greater
 import collections
 import itertools
 import pickle
@@ -59,6 +61,9 @@ def _atom_slice(traj, indices):
     if traj._have_unitcell:
         unitcell_lengths = traj._unitcell_lengths.copy()
         unitcell_angles = traj._unitcell_angles.copy()
+    else:
+        unitcell_lengths = None
+        unitcell_angles = None
     time = traj._time.copy()
 
     # Hackish to make the smart slicing work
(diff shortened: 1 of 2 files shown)


sroet:
Looks good to me as well.

If it is not to much to ask, I would really like to also see a test. Just to make sure that this behavior does not regress.

nffaruk:
> One question: was there anything specific about your use case that led to this problem? (other than using a square aspect ratio?) I'm asking so that, in case we ever come back to revisit this approach, we can be sure not to break what you've fixed!

Thanks for the approval - in my case my research group has it's own CG MD software that does not have solvent and thus no unit cells, and so `return md.Trajectory(...)` in `def _atom_slice(...)` was throwing `UnboundLocalError: local variable 'unitcell_lengths' referenced before assignment` - the unit cell kwargs are set to None by default in the mdtraj API for the Trajectory class [by the way](http://mdtraj.org/1.9.3/api/generated/mdtraj.Trajectory.html#mdtraj.Trajectory).

A more general case where others might run into this includes PDB files with dummy CRYST1 records (which mdtraj recognizes and removes) or ones that don't have any such records (eg. curated benchmark sets for folding and docking).

> If it is not to much to ask, I would really like to also see a test. Just to make sure that this behavior does not regress.

I think it would be as simple as making a copy of the .../contact_map/contact_map/tests/trajectory.pdb without the CRYST1 lines. The calculated contacts may not be the same as before because mdtraj.compute_neighbors(...) will not treat it as periodic.

sroet:
Written a test that confirms this fixes the described error #65. Merging this in.

Thank you very much for the contribution!

PR #65, opened 2019-11-17
Added test for non_periodic systems

sroet:
This adds an unittest for the failing behavior described by #64 
Fails on current master before that MR was merged, passes on that branch, should pass on the current master

(Also does some minor pep8 fixes)

### DIFF
--- a/contact_map/tests/test_contact_map.py
+++ b/contact_map/tests/test_contact_map.py
@@ -1,6 +1,7 @@
 import os
 import collections
 import mdtraj as md
+import copy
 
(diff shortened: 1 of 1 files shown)


dwhswenson:
LGTM, and passes when re-running after the merge of #64.

On some of the style changes -- I usually prefer extra indentation when using backslash continuation. PEP8 doesn't have an opinion on this, but suggests extra indentation as one option. I think this may be one of the areas where I differ from `black`. Anyway, doesn't really matter -- I think it only shows up in a few `assert`s.

PR #66, opened 2019-12-06
Move to Autorelease 0.1.0

dwhswenson:
This PR updates to use the new version [Autorelease](https://github.com/dwhswenson/autorelease) (my framework for handling releases). I've already done releases with the new Autorelease for 2 other projects (Autorelease itself being one of them), so it should work.

The new version has a number of advantages. Here are a couple:

* It uses the new Travis config import system, so all the fancy Travis stages are maintained in the Autorelease repository (note the simplification of the `.travis.yml`!)
* It has better handling of versions (correct version number updates in developer installs)
* Previously, if an error occurred after the deployment to testpypi, the only solution was to bump the version, leading to version skips. This is no longer the case -- you only need to bump the version if the error occurs after releasing on GitHub. (The trick to do this was a little fancy: I rewrite the setup.cfg to say that it is a development version, and increase the dev release number depending on what is already on testpypi. So testpypi only has dev versions.)

One major change is that all project-specific information is now in `setup.cfg`. The `setup.py` is project-independent, and vendored from Autorelease. The `version.py` is also vendored from Autorelease. When we want to upgrade to a newer version of Autorelease, we can copy the relevant files in (actually, by then Autorelease might have helper scripts for that, since that's a likely target for Autorelease 0.2.0).


dwhswenson:
@sroet : There's not much to do for reviewing this. All the code is vendored, so any changes to it should be made in the original location (in autorelease). Testing should also be there. Mainly asking for review so that you're aware of the change in structure around these parts of the project.

contact_map/version.py
```diff
@@ -0,0 +1,141 @@
+# This file vendored from Autorelease
+import os
+import subprocess
+
+try:
+    from configparser import ConfigParser, NoSectionError, NoOptionError
+except ImportError:
+    # py2
```
sroet: small question, do we really want to keep support for py2 around (EOL is approaching)

dwhswenson: Vendored code. Autorelease supports 2.7 because it is used (well, about to be used) in other projects that do support 2.7 (like the upcoming 1.x releases of OPS).

Having the support there makes no difference even if (when) Contact Map Explorer goes all-Py3. It just won't be used. Since it is vendored, it doesn't count as something that we have to maintain here. (I have to maintain it *elsewhere*, not here!)

This does get back to the conceptual thing that's kind of weird here: `setup.py` and `version.py` should be considered read-only (as much as possible). If they're changed, the changes should happen in Autorelease.

Normally I'm very opposed to vendoring code, because of exactly this problem of synchronizing changes. Better to make it a requirement. However, when the `setup.py` itself is what is vendored, you end up in the trade-offs that come with PEP517 (including losing developer installs). Vendoring seemed like the lesser evil.

sroet: > Vendoring seemed like the lesser evil.

I agree with you, and could not think of a better way to resolve this at the moment. Feel free to merge on green

dwhswenson:
AppVeyor is failing on ARCH=64 because for some reason it isn't copying the stuff from MANIFEST.in. This is weird, and will take some looking into.

Issue #67, opened 2020-01-31
UnboundLocalError: local variable 'unitcell_lengths' referenced before assignment


Jiaxing2018:
I was trying to generate the contact frequency map using my .xtc and .pdb file. However, it has this error. I used the .xtc and .pdb files in the example and it works well. I don't know if there is anything wrong with my .xtc or .pdb file? Thank you for your help!

Jiaxing2018:
UnboundLocalError                         Traceback (most recent call last)
<ipython-input-15-34460bccab7f> in <module>
      8 
      9 
---> 10 trajectory_contacts = ContactFrequency(traj)
     11 
     12 # if you want to save this for later analysis

~/anaconda/envs/jupyter_notebook/lib/python3.6/site-packages/contact_map/contact_map.py in __init__(self, trajectory, query, haystack, cutoff, n_neighbors_ignored, frames)
    741                                                query, haystack, cutoff,
    742                                                n_neighbors_ignored)
--> 743         contacts = self._build_contact_map(trajectory)
    744         (self._atom_contacts, self._residue_contacts) = contacts
    745 

~/anaconda/envs/jupyter_notebook/lib/python3.6/site-packages/contact_map/contact_map.py in _build_contact_map(self, trajectory)
    777         residue_query_atom_idxs = self.residue_query_atom_idxs
    778 
--> 779         used_trajectory = self.slice_trajectory(trajectory)
    780         for frame_num in self.frames:
    781             frame_contacts = self.contact_map(used_trajectory, frame_num,

~/anaconda/envs/jupyter_notebook/lib/python3.6/site-packages/contact_map/contact_map.py in slice_trajectory(self, trajectory)
    585         if self.use_atom_slice and (len(self._all_atoms) <
    586                                     trajectory.topology.n_atoms):
--> 587             sliced_trajectory = _atom_slice(trajectory, self._all_atoms)
    588         else:
    589             sliced_trajectory = trajectory

~/anaconda/envs/jupyter_notebook/lib/python3.6/site-packages/contact_map/contact_map.py in _atom_slice(traj, indices)
     68     topology._numAtoms = len(indices)
     69     return md.Trajectory(xyz=xyz, topology=topology, time=time,
---> 70                          unitcell_lengths=unitcell_lengths,
     71                          unitcell_angles=unitcell_angles)
     72 

UnboundLocalError: local variable 'unitcell_lengths' referenced before assignment

sroet:
[...]

PR #69, opened 2020-02-08
Fixes for sparse matrix in Pandas 1.0

dwhswenson:
This gets the tests running under pandas 1.0, but they fail with an assertion error. I think (still need to confirm) that this is the same as https://github.com/pandas-dev/pandas/issues/29814. If so, then it is a bug in pandas.

Maybe release 0.5.0 with a pin to `pandas<1`? It doesn't look like it is easy to generate sparse matrices directly from dict-of-keys in pandas; pandas seems to prefer to take a dense matrix and make it sparse. We don't want to create the dense matrix in the first place.

### DIFF
--- a/ci/conda-recipe/meta.yaml
+++ b/ci/conda-recipe/meta.yaml
@@ -20,7 +20,7 @@ requirements:
     - mdtraj
     - numpy
     - scipy
-    - pandas<1.0
+    - pandas
 
 test:
   requires:
--- a/contact_map/contact_count.py
+++ b/contact_map/contact_count.py
@@ -25,6 +25,33 @@ def _colorbar(with_colorbar, cmap_f, norm, min_val):
     return cb
 
 
+# TODO: remove following: this is a monkeypatch for a bug in pandas
+# see: https://github.com/pandas-dev/pandas/issues/29814
+from pandas._libs.sparse import BlockIndex, IntIndex, SparseIndex
+def _patch_from_spmatrix(cls, data):
+    length, ncol = data.shape
+
+    if ncol != 1:
+        raise ValueError("'data' must have a single column, not '{}'".format(ncol))
+
+    # our sparse index classes require that the positions be strictly
+    # increasing. So we need to sort loc, and arr accordingly.
+    arr = data.data
+    #idx, _ = data.nonzero()
+    idx = data.indices
+    loc = np.argsort(idx)
+    arr = arr.take(loc)
+    idx.sort()
+
+    zero = np.array(0, dtype=arr.dtype).item()
+    dtype = pd.SparseDtype(arr.dtype, zero)
+    index = IntIndex(length, idx)
+
+    return cls._simple_new(arr, index, dtype)
+
+pd.core.arrays.SparseArray.from_spmatrix = classmethod(_patch_from_spmatrix)
+# TODO: this is the end of what to remove when pandas is fixed
+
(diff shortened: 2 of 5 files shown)


sroet:
I probably will not have time to look at this in detail this week.
However, a quick google did give me a [way of migrating](https://pandas.pydata.org/pandas-docs/stable/user_guide/sparse.html#migrating) with [Scipy Sparse matrices](https://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.sparse.dok_matrix.html)

dwhswenson:
@sroet : That's the problem. We currently go from scipy to pandas. I've gone through the steps in that guide (that's why the code runs now with an error in correctness, not an error in running). The bug in pandas-dev/pandas#29814 is that pandas isn't reading correctly from `scipy.sparse` (or possibly, just that `pandas.sparse` has a fundamental error in it). In either case, I think it's a problem that should be fixed on the pandas side; I'm also not currently at a point where I have time to make a PR to fix it there.

Still haven't verified that this is definitely the same problem as that, but it seems pretty likely, since it is an unresolved pandas bug with making sparse DFs from scipy matrices.

Main question then is, do we release 0.5.0 pinned to an older version of pandas? (i.e., without this PR). Since it seems like we're both pretty busy, I lean toward pinning pandas and releasing 0.5.0 to avoid more users having problems like #67.

sroet:
> Main question then is, do we release 0.5.0 pinned to an older version of pandas? (i.e., without this PR). Since it seems like we're both pretty busy, I lean toward pinning pandas and releasing 0.5.0 to avoid more users having problems like #67.

+1 on this

dwhswenson:
I've got a monkey patch working for this, but I want to see if it can get fixed quickly in pandas instead.

In any case, supporting py27 is now definitely too cumbersome. We'll release this as a 0.5.1 (hopefully without needing to monkeypatch pandas), but then let's move master to 0.6.0.dev0, dropping py27. I also vote for dropping win_32 testing from AppVeyor, instead using highest/lowest Python versions (giving 2 tests instead of 4; free AppVeyor doesn't run in parallel). Our AppVeyor testing is based on MDTraj, and probably MDTraj of 7 years ago.

sroet:
LGTM, feel free to merge at your discretion (if pandas fix takes to long)

> In any case, supporting py27 is now definitely too cumbersome. We'll release this as a 0.5.1 (hopefully without needing to monkeypatch pandas), but then let's move master to 0.6.0.dev0, dropping py27. I also vote for dropping win_32 testing from AppVeyor, instead using highest/lowest Python versions (giving 2 tests instead of 4; free AppVeyor doesn't run in parallel).

+1 for dropping py27 and win_32

dwhswenson:
It looks like this problem may have been resolved by https://github.com/pandas-dev/pandas/pull/32825 (although I can't seem to get a working dev build of pandas going to verify it.) That's a complete rewrite of `DataFrame.sparse.from_spmatrix` that bypasses the problematic code in `SparseArray.from_spmatrix`. I think I'll merge this for now, and sometime after pandas 1.1 is released (with that fix) we can require pandas>=1.1 and remove the patch. The good thing is that the patch won't even be called once pandas 1.1 is out, since the buggy code is bypassed in the new implementation.

I will also release 0.5.1, which will mean I stop getting emails about our stable branch failing on Travis because MDTraj-dev no longer supports Py2 (see #75).

PR #72, opened 2020-02-14
Fix RTD docs build

dwhswenson:
For a few months, docs on RTD have been failing because the conda-based install is too memory-intensive. This switches back to a setuptools/pip-based install. I think I'd switched to conda when there was some problem installing MDTraj over pip; now I switch back.

I don't think there were any changes to docs between last successful build of `master` (`latest` in docs) and now, except the date bump in this PR. However, the docs for `stable` still reflect 0.4.0, not 0.5.0 (name and credits are the giveaways).

This PR fixes it; you can verify that in the RTD build of the docs branch (with updated copyright year): https://contact-map.readthedocs.io/en/docs/


sroet:
LGTM, feel free to merge on green.

PR #75, opened 2020-06-30
Drop Py27 and Win32; add Py38

dwhswenson:
Currently our Travis script has a Py27/MDTraj dev test job. However, MDTraj has made changes that are incompatible with Py27. We'd already discussed dropping Py27 and Win32 from tests (https://github.com/dwhswenson/contact_map/pull/69#issuecomment-585827472), so I'm doing it now.

While I'm at it, I'm including Py38 in our test matrix (overdue!). Note that the next release will be the last one to guarantee support Python 3.6 (by NEP29 we could already drop 3.6, but I'm going to take a slightly more conservative approach.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,8 @@ version: ~> 1.0
 
 language: python
 python:
+  - '3.8'
   - '3.7'
-  - '2.7'
   - '3.6'
 
 notifications:
@@ -17,7 +17,7 @@ branches:
 
 env:
   global:
-    - CANONICAL_PYTHON="3.7"
+    - CANONICAL_PYTHON="3.8"
     - CANONICAL_MDTRAJ="release"
     - CODECLIMATE=""
     - TWINE_USERNAME="dwhswenson"
@@ -33,6 +33,8 @@ jobs:
   exclude:
     - env: MDTRAJ="dev"
       python: "3.6"
+    - env: MDTRAJ="dev"
+      python: "3.7"
 
 before_install:
   - echo "before install"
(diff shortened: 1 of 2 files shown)


dwhswenson:
Going to close/open to see if I can get travis to trigger

dwhswenson:
Summary of what's in here:

* We no longer test against Python 2 at all
* We no longer test against Win32 at all
* Travis tests Python 3.6, 3.7, 3.8, and considers 3.8 the "canonical" version (and 3.8 is tested against MDTraj current master as well as MDTraj release; others only tested against MDTraj release)
* AppVeyor tests Python 3.6, 3.7. (It seems that no one is using AppVeyor's Miniconda38 yet; AppVeyor lists it as deployed, but I couldn't find any evidence of widespread use)

Also, I migrated us to travis-ci.com (from the .org). There's a general (but very gentle) push from the Travis folks to go that direction.

As a thought for the future, it might make sense for testing against MDTraj master to be skipped on the `stable` branch. But that's for a future PR.

appveyor.yml
```diff
@@ -14,26 +14,24 @@ environment:
...
+    - PYTHON: "C:\\Miniconda36-x64"
+      CONDA_PY: "36" 
+      ARCH: '64'
 
 build: false
 
 install:
   # install python
-  - SET PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%
+  - SET PATH=%PYTHON%;%PYTHON%\\Scripts;%PYTHON\\bin;%PATH\\Library\\bin;%PATH%
```
sroet, line 31: Why are the extra PATH variables needed? (just asking for info)

appveyor.yml
```diff
@@ -14,26 +14,24 @@ environment:
...
   - conda config --add channels conda-forge
-  - conda install -yq conda=4.5 conda-build=3.10
+  #- conda install -yq conda=4.5 conda-build=3.10
```
sroet, line 33: please remove the old (I presume temporary) install command. Or do you think we might need it in the future?

.travis.yml
```diff
@@ -33,6 +33,8 @@ jobs:
   exclude:
     - env: MDTRAJ="dev"
       python: "3.6"
+    - env: MDTRAJ="dev"
+      python: "3.7"
```
sroet, line 37: Wouldn't a single `include` (MDTRAJ="dev", python="3.8") be better than 2 `exclude`s? Just a mater of opinion, feel free to ignore.

appveyor.yml
dwhswenson: removed; but may be required in the future. Newer versions of conda have moved the conda executable around, so if we're manually setting `PATH`, it might have moved. I added these while searching for it in the supposed-to-exist Miniconda38 on AppVeyor.

appveyor.yml
dwhswenson: done

.travis.yml
dwhswenson: I agreed, and tried it, and the build matrix was WTF weird. See https://travis-ci.com/github/dwhswenson/contact_map/builds/174036372.

Build 2313.4, which says it is deploy testpypi, is actually the extra test with `MDTRAJ="dev"`. So it does what we want, but looks too weird. I'm too busy with other stuff to raise an issue with the Travis people right now, but it might be a bug in how build stages plays with explicit includes. At the least, I find the behavior unexpected. Could be related to different "merging" styles that Travis uses, too.

sroet: huh, that is very weird indeed, fine by me to leave it for now

PR #79, opened 2020-07-13
rework plotting to allow for plot_axes

### LINKED ISSUE #78: Subplots for ContactFrequency
Hello again, how would we go about arranging the plots generated through the plot method of ContactCount for different trajectories as subplots? I've tried [this Stack answer](https://stackoverflow.com/a/46906599) without any luck. Perhaps there should be an option to provide an existing figure or axis object to the plotting method? Thank you,

sroet:
Closes #78 

This reworks the plotting of `ContactCount`, it split out the axes plotting code in `ContactCount.plot_axes()`

This also allows users to use `plot_axes` to allow for greater flexibility (like with `subplots`)

This now works:
```python
import contact_map as cm
import mdtraj as md
import matplotlib.pyplot as plt

traj = md.load("contact_map/tests/trajectory.pdb")
contactmap = cm.ContactFrequency(traj)
# Old style
fig1, ax1 = contactmap.atom_contacts.plot(figsize=(12,12))

# New posibility
fig, axs = plt.subplots(2,2, figsize=(12,12))
axs = axs.flatten()
for ax in axs:
    _ = contactmap.atom_contacts.plot_axes(ax=ax)
```

~BIG DIFFERENCE between `plot()` and `plot_axes()`: No `colorbar` for `plot_axes()`~

~Colorbars are hard for subplots in `matplotlib` (part of the figure, but also plotted with the last `ax` of `subplots`)
An user can add a seperate colorbar to the figure if they want, using any of the answers [here](https://stackoverflow.com/questions/13784201/matplotlib-2-subplots-1-colorbar).~

Edit: current default plots a colorbar in every axes, can be turned off `with_colorbar=False`

[API-BREAK] - in `.plot_utils` the call-signature of `ranged_colorbar` has changed: 
    - unused kwarg `name` is removed
    - new kwarg `ax` is added, this is the `axes` that is put into `plt.colorbar( ax=ax)` 

@nffaruk Would you be able to see if this solves your issue?
You can install this version of the code by running
`pip install git+https://github.com/sroet/contact_map.git@allow_fig_ax_kw` 
preferably in a new `conda` environment (or you might need to uninstall first, depending on how you installed Contact Map Explorer in the first place)


dwhswenson:
Both of us putting thought into this today! Posts with minutes of each other. 😆 

For colorbars, try passing `ax` down through `contact_counter._colorbar` and to `ranged_colorbar`. Then something like:

```python
if ax is None:
    colorbar = plt.colorbar
else:
    # might actually need a partial here, if plt.colorbar doesn't have ax/cax params
    colorbar = ax.figure
cb = colorbar(sm, fraction=0.046, pad=0.04, ax=ax)  # maybe cax=ax?
```

See [docs on `Figure.colorbar`](https://matplotlib.org/api/_as_gen/matplotlib.figure.Figure.html#matplotlib.figure.Figure.colorbar) for details on `ax` and `cax` kwargs.

dwhswenson:
The idea above should act as I suggested in #78: defaults to adding a colorbar for each axis, although the user can override with `with_colorbar=False`. Then the next step would be to make a more accessible version of `colorbar`. I think the signature for that would be something like:

```python
def colorbar(cmap, is_divergent, cbar_min, ax=None):
    ...
```

`cmap` could be string (`plt.get_cmap` to get the func) or an actual cmap function. `is_divergent` is a bool, and would use `vmin, vmax = -1.0, 1.0` if True, `0.0, 1.0` if False. `cbar_min` is `0` or `-1`. If you look at `contact_counter._colorbar`, the main thing it does is use `np.floor` to map `-1.0 ≤ x < 0.0` to `cbmin = -1` and `0.0 ≤ x < 1.0` to `cbmin = 0`.

I'll need to write some docs at some point to explain how to use divergent vs. sequential color maps, and why I think there's so much virtue in using the top half of a divergent color map. But this is definitely a more user-friendly way to build a colorbar.

The point of all of this is to prep info that gets fed into `ranged_colorbar`, so it can actually create the colorbars. With this, passing the `ax` with multiple axes (after using `with_axes=False` in the original call to `plot_axes()`) should let you automatically create a single colorbar for the whole figure. (again, see the docs on `ax` in `Figure.colorbar`.)

I also lean toward the overall idea of gathering some of the matplotlib stuff together at some point. In principle, the future might involve non-matplotlib plotting (e.g., interactive Bokeh that tells you more about the contact when you hover). Feel free to move things around if it makes sense to you.

nffaruk:
> @nffaruk Would you be able to see if this solves your issue?
> You can install this version of the code by running
> `pip install git+https://github.com/sroet/contact_map.git@allow_fig_ax_kw`

Thanks both of you for looking into this in detail, and so quickly too! Unfortunately, I'm unable to install that version, I'm still on Python 2.7.16 (for compatibility with my lab's software) and I get this error:

```
    ERROR: Command errored out with exit status 1:
     command: /home/nffaruk/.conda/envs/upside/bin/python -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-req-build-Xt1kBk/setup.py'"'"'; __file__='"'"'/tmp/pip-req-build-Xt1kBk/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-pip-egg-info-1Jci8s
         cwd: /tmp/pip-req-build-Xt1kBk/
    Complete output (12 lines):
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-req-build-Xt1kBk/setup.py", line 126, in <module>
        write_installed_version_py()
      File "/tmp/pip-req-build-Xt1kBk/setup.py", line 99, in write_installed_version_py
        version_finder = VersionPyFinder()
      File "/tmp/pip-req-build-Xt1kBk/setup.py", line 29, in __init__
        self.functions = self._get_functions(self.filename)
      File "/tmp/pip-req-build-Xt1kBk/setup.py", line 93, in _get_functions
        exec(compile(tree, filename="version.py", mode='exec'), locs)
      File "version.py", line 12, in <module>
    TypeError: exceptions must be old-style classes or derived from BaseException, not NoneType
    ----------------------------------------
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
```

nffaruk:
Also, regarding #78 @dwhswenson 's option 3 with the kwargs, [according to this](https://stackoverflow.com/a/15302042), it doesn't seem like there would there be an issue mixing explicitly named/default kwargs with the arbitrary **kwargs unpacking?

sroet:
@nffaruk Unfortunately, [`Python 2` has reached end-of-life on 01-01-2020](https://www.python.org/doc/sunset-python-2/).

I see you are using conda, is there a particular reason why you can't separate the conda environments for your lab software and contact map explorer?

If not you can try the following:
```
conda create -n contact_map python=3
conda activate contact_map 
conda install --only-deps contact_map
pip install git+https://github.com/sroet/contact_map.git@allow_fig_ax_kw
```
This will make a new conda environment for contact map, if you want to activate it in a new shell you can use `conda activate contact_map`

> Also, regarding #78 @dwhswenson 's option 3 with the kwargs, according to this, it doesn't seem like there would be an issue mixing explicitly named/default kwargs with the arbitrary **kwargs unpacking?

The problem @dwhswenson indicated is that not all valid `**kwargs` for  `matplotlib.subplots()` lead to error free plotting code on our side (i.example if you give `nrows=2` or any other way that we generate more than 1 `axes`, it fails, as we don't know which axis to plot in).

dwhswenson:
@sroet : This particular problem can be fixed by updating the vendored copy of `setup.py`. I had to fix this problem for OPS a few months back. Diff: https://github.com/dwhswenson/autorelease/commit/4e74069e649487462d50bd51d9e0accd9c8ae388.

I don't think we've added anything that breaks Py2 compatibility yet, but that will definitely come soon.

sroet:
@dwhswenson any preference on the codeclimate issue? (Function ranged_colorbar has 6 arguments (exceeds 5 allowed) 

I opted to remove the `name` `kwarg`, as it was unused/hard-coded anyway

dwhswenson:
> @dwhswenson any preference on the codeclimate issue? (Function ranged_colorbar has 6 arguments (exceeds 5 allowed)
> 
> I opted to remove the name kwarg, as it was unused/hard-coded anyway

That's fine. Or I'll ignore it, if codeclimate keeps complaining. It is designed to be a function that I can copy into other projects in the future. It does a lot more than what we use.

sroet:
> @sroet : This particular problem can be fixed by updating the vendored copy of `setup.py`. I had to fix this problem for OPS a few months back. Diff: [dwhswenson/autorelease@4e74069](https://github.com/dwhswenson/autorelease/commit/4e74069e649487462d50bd51d9e0accd9c8ae388).
> 
> I don't think we've added anything that breaks Py2 compatibility yet, but that will definitely come soon.

awesome, applied the same code. @nffaruk do you mind rerunning the install?

sroet:
@dwhswenson this is ready for a review (apparently our current test-suite catches all changes made to the code in this PR)

nffaruk:
Got it to install and it works great, thanks! For sure developers shouldn't have the burden of supporting end of life software and it's straightforward to switch environments, but I was hoping there would be an easy fix considering our PI is a big user of your package (this subplot request came from him) and uses it in conjunction with a bunch of analysis functions from our side. But we will have to get on the inescapable migration to Python 3 soon.

sroet:
>  I noticed that the colorbars don't go the full height of the plot

Yeah, I think this is a difference between "taking space from an existing axes" and "adding a colorbar axes", but not to sure about that. Give me an hour to test something out

dwhswenson:
No rush... I probably won't merge this until after 0.5.1 gets released, anyway. I have a few ideas to test that might fix the release process.

sroet:
Figured it out, it is due to the changed aspect ratio of the axis (default is 10,8) but for that plot I used (20,6). If I change it to (30,8) it plots as intended again.

sroet:
All set ;)

dwhswenson:
With 0.5.1 out, I'll go ahead and merge this as the first step to 0.6.0!

PR #80, opened 2020-08-12
Bump 0.6.0.dev0

dwhswenson:
"Officially" this is the version-bump to 0.6.0.dev0, although that technically already happened in #79. But we need to merge stable -> master in order to avoid git conflicts on the version line, plus the 0.5.1 release branch bumped up the version of autorelease that we're using, so that should also come back to master.


sroet:
LGTM feel free to merge

Issue #53, opened 2019-04-16
Enhancement: ContactTrajectory class (with movies!)


dwhswenson:
It might be nice to add a `ContactTrajectory` class. This would store a separate (single frame) contact map for each time slice. Then you could do interesting things like create a `ContactFrequency` based on a set of frames determined after the fact. You could also create movies of how the contact map evolves in time (these should be based on some windowing into `ContactFrequency` instances; the window can be 1 if the user wants).

Relevant matplotlib docs on animation: 
* https://matplotlib.org/api/animation_api.html
* https://matplotlib.org/gallery/animation/dynamic_image.html

Animations that could be generated in real time would be impressive, but I'd be just as happy with something that pre-generates the images and makes a movie out of them, like `ArtistAnimation` does.

Note that, to use a rolling window to generate the contact frequency plots for each movie frame, we already have `add_contact_frequency` and `subtract_contact_frequency` methods. These affect the absolute histogram count, unlike `ContactDifference`, which compares the after-normalization (to number of frames) contact frequencies. We'd still output a separate plot at each time step, but the underlying data would be cheaper to calculate.

sroet:
+1 for the idea.

>We'd still output a separate plot at each time step, but the underlying data would be cheaper to calculate.

Which one do you expect to be cheaper, using the methods or the `ContactDifference`?

dwhswenson:
> Which one do you expect to be cheaper, using the methods or the `ContactDifference`?

Not sure what you mean here... these calculate different things.

If you have `freq_1` where a given contact is present 20/25 frames (80%), and `freq_2` where that contact is present 1/2 frames (50%), then `ContactDifference(freq_1, freq_2)` gives a result of 30% for that contact (80% - 50%). On the other hand, `freq_1.subtract_contact_frequency(freq_2)` results in a `ContactFrequency` where the contact is present in 19/23 frames (~83%). Of course, this only makes sense if you know that the trajectory that makes `freq_2` is a subtrajectory of the one that made `freq_1`.

Those methods were designed specifically for doing this sort of windowing of a contact frequency, so those are the things to use.

sroet:
>Those methods were designed specifically for doing this sort of windowing of a contact frequency, so those are the things to use.

Ah I understand now :) thanks for the explanation

sroet:
@dwhswenson was there any issue left here after #81 ?

dwhswenson:
#81 doesn't include movies, just the `ContactTrajectory` itself.

[...]

PR #81, opened 2020-08-13
Contact trajectories

dwhswenson:
The main purpose of this is to introduce a `ContactTrajectory` subclass of `ContactObject`. It also comes with a few related additions and tools. Major step toward #53.

### `ContactTrajectory`

This is essentially a list of single-frame `ContactFrequency` instances. A contact map for a specific frame can be obtained by either:

```python
# these are the same
contacts = contact_trajectory[frame_num].residue_contacts
contacts = contact_trajectory.residue_contacts[frame_num]
```

The normal `ContactTrajectory` is a `Sequence`; there is also a `MutableContactTrajectory` which is a `MutableSequence`.

This PR will not include the movie making aspects suggested in #53 -- that will be left for future work.

### `ContactObject.from_contacts`

For efficiency, the easiest way to calculate the contact trajectory is to have a loop that calculates the contacts for each frame, and then to create a `ContactFrequency` from `atom_contacts` and `residue_contacts`. We did not have a method to create a `ContactFrequency` from contacts before, so I added it. Like many things, it isn't allowed for `ContactDifference`, where it is overridden with a `NotImplementedError`.

The input to `ContactObject.from_contacts` can be either `ContactCount` instances, or the underlying `collections.Counter`s (as well as, of course, the necessary information for `ContactObject.__init__`).

### `WindowedIterator` and `ContactTrajectory`-specific implementation

The basic idea here is to facilitate rolling-average type behavior (perhaps in our case, rolling-frequency). By iterating over an underlying sequence, the idea is to return sequence indices that should be added to/subtracted from the rolling average. There's a general implementation (perhaps useful in other code!) and a specific use of it to get rolling contact frequencies.

---

### Tasks

- [x] Implement and test `ContactObject.from_contacts`
- [x] Implement core of `ContactTrajectory`
- [x] Test core of `ContactTrajectory`
- [x] Test with sliced trajectory
- [x] Add compatibility checks where relevant in `ContactTrajectory`, with tests
- [x] Implement `MutableContactTrajectory`, with tests
- [x] `WindowedIterator` and tests
- [x] `RollingContactFrequency` and tests
- [x] Docstrings
- [x] Add example
- [x] Add to docs


dwhswenson:
Quick comments on the codeclimate complaints:

* `from_contacts` has too many arguments (3 times): True, but unavoidable. And I'll also make the arguments that, (1) most of those arguments typically use defaults; (2) while this is part of the API, it's aimed less at normal users than at developers. Both of these make the too-many-arguments thing less of a problem in my mind.
* file `contact_maps.py` is too long: Yes. I'm not going to change that in this PR, but I was already thinking about reorganizing some of the code here. That will be a separate PR.
* TODO: yes, I added a TODO... that shouldn't be blocking, though

I'll silence those issues for now: `wontfix` on the too many arguments, `confirmed` on the too long file. Approving the changes.

dwhswenson:
This is ready for review. Once this one is in, I'll cut 0.6.0.

contact_map/contact_trajectory.py
```diff
@@ -0,0 +1,368 @@
...
+    Internally, this has a single-frame :class:`.ContactFrequency` for each
+    frame of the trajectory.
+
+    Parameters
+    ----------
+    trajectory : mdtraj.Trajectory
+        the trajectory to calculate contacts for
+    query : list of int
+        Indices of the atoms to be included as query. Default ``None``
+        means all atoms.
```
sroet, line 18: this is heavy atoms and not water (from `ContactObject`)
We might want to change that default
```suggestion
        means all heavy, non-water atoms.
```

contact_map/contact_trajectory.py
```diff
@@ -0,0 +1,368 @@
...
+    haystack : list of int
+        Indices of the atoms to be included as haystack. Default ``None``
+        means all atoms.
```
sroet, line 21: same as above
```suggestion
        means all heavy, non-water atoms.
```

contact_map/contact_trajectory.py
```diff
@@ -0,0 +1,368 @@
...
```
sroet, line 209: Same as above
```suggestion
        means all heavy, non-water atoms.
```

contact_map/contact_trajectory.py
```diff
@@ -0,0 +1,368 @@
...
+class MutableContactTrajectory(ContactTrajectory, abc.MutableSequence):
+    """Mutable version of :class:`.ContactTrajectory`
+
+    Parameters
+    ----------
+    trajectory : mdtraj.Trajectory
+        the trajectory to calculate contacts for
+    query : list of int
+        Indices of the atoms to be included as query. Default ``None``
+        means all atoms.
```
sroet, line 206: same as above:
```suggestion
        means all heavy, non-water atoms.
```

contact_map/tests/test_contact_trajectory.py
```diff
@@ -0,0 +1,350 @@
...
+class TestMutableContactTrajectory(object):
+    def setup(self):
+        self.traj = md.load(find_testfile("trajectory.pdb"))
+        self.map = MutableContactTrajectory(self.traj, cutoff=0.075,
+                                            n_neighbors_ignored=0)
+        self.expected_atom_contacts = TRAJ_ATOM_CONTACTS.copy()
+        self.expected_residue_contacts = TRAJ_RES_CONTACTS.copy()
+
+    def _test_expected_contacts(self, traj_map, exp_atoms, exp_res):
+        i = 0  # debug
```
sroet, line 186: leftover debug statement?
```suggestion

```

contact_map/tests/test_contact_trajectory.py
```diff
@@ -0,0 +1,350 @@
...
+        for cmap, exp_a, exp_r in zip(traj_map, exp_atoms, exp_res):
+            atom_counter = cmap.atom_contacts.counter
+            res_counter = cmap.residue_contacts.counter
+            assert atom_counter == counter_of_inner_list(exp_a)
+            assert res_counter == counter_of_inner_list(exp_r)
+            i += 1
```
sroet, line 192: leftover debug statement?
```suggestion

```

sroet (changes requested):
LGTM codewise, some doc changes and leftover debug code in the tests

contact_map/contact_trajectory.py
dwhswenson: I think I took this as default because MDTraj had something similar. One thing: if the default atoms change, the default cutoff should also change.

dwhswenson:
Made changes as suggested, @sroet. Thanks for the quick review!

sroet (approved):
LGTM, feel free to merge on green

PR #83, opened 2020-08-13
Coveralls cleanup

dwhswenson:
Coveralls has been (silently) failing since coverage 5 was released (late 2019). This brings coveralls back to our PRs (by switching from the defunct `python-coveralls` to `coveralls`) and excludes from coverage a few things that are no longer expected to be covered by tests (stuff for Python 2 and pandas < 1.1).

The things that are currently excluded should, at some point, be removed. But I see no reason to rush to break things that work for some users -- besides, we haven't officially updated to require these more recent versions; just we don't guarantee that the old versions will continue to work.

### DIFF
--- a/ci/pip-install/testing_requirements.txt
+++ b/ci/pip-install/testing_requirements.txt
@@ -1,5 +1,5 @@
 pytest
 pytest-cov
-python-coveralls
+coveralls
 codacy-coverage
 autorelease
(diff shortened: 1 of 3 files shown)


dwhswenson:
Ready for review. With this, it looks like we still have one genuine missed line, but I feel pretty good about 99.853% coverage. And I feel much better to actually see the coverage reports in PRs.

Issue #82, opened 2020-08-13
Renaming `ContactFrequency` to `ContactMap`; removing old `ContactMap`


dwhswenson:
Currently, Contact Map Explorer has classes `ContactMap` and `ContactFrequency`. Really, a `ContactMap` is nearly indistiguishable from a `ContactFrequency` with only one frame -- at least, I don't think there's anything `ContactMap` does that `ContactFrequency` doesn't -- so it seems logical to remove the extra code.

Here's my proposed process for doing that:

* 0.6.0: Deprecate `ContactMap`
* 0.7.0: Remove `ContactMap`; leave a function that gives an error saying to use `ContactFrequency`. Also, FutureDeprecationWarning on `Contact Frequency` for the name change to `ContactMap`
* 0.8.0 or 1.0.0: Deprecate `ContactFrequency`, with the `ContactMap` name preferred. Leave a deprecated `ContactFrequency` until 2.0 (which may be a *very* long time).

Each release will also include other features, but I'm going to encourage a fast release cycle here (add a new feature-level change? good reason for a new release!) Unfortunately, that also means that even if this is a reasonable timeline in terms of version numbers, it'll be a fast timeline in terms of months. (I hope.)

Thoughts on this? I plan to pin this issue for a while well after the rename is complete, so any confused users can easily find it.

sroet:
Only addition is that I would also put it on the release notes for each of these steps. Other than that LGTM

dwhswenson:
Yes, I'll add `deprecation` and `api break` to the labels we can assign to PRs. This will ensure that the automated draft of the release notes highlights these changes, reminding me to add that to the summary.

PR #84, opened 2020-08-14
Deprecate `ContactMap` (allowing rename `ContactFrequency` => `ContactMap`)

dwhswenson:
First step of #82.

Also, note that the correct warning to use here is `FutureWarning`, which will actually show for users (`DeprecationWarning` will not). I learned that from [this SO answer](https://stackoverflow.com/a/55378483/4205735), written by a very brilliant author. I would argue that basic interactive use of Contact Map Explorer in a Jupyter notebook counts as "application", not "library."


dwhswenson:
Codeclimate complains about:

* contact_map.py being too long (well, it'll be shorter when we remove `ContactMap`! plus, as stated before, I intend to split things out of it in a PR dedicated to that process)
* the changes for deprecating classmethods look like repeated code. While writing the code, I actually thought about writing a decorator to avoid that repetition, but it seemed like overkill for code that's going to be removed in a near-future release.

Marking both as wontfix.

contact_map/contact_map.py
```diff
@@ -7,6 +7,9 @@
 import itertools
 import pickle
 import json
+
+import warnings
+
```
sroet, line 12: Any reason for the excess newlines around this import?

dwhswenson: reminder that this is temporary and should be removed

dwhswenson:
> * in the contact_map.ipynb there are still a couple references to ContactMap (an import and some text in the `list residue contacts` section

Fixed.

> * The without_atom_slice notebook still references contactMap (as the only bad case), so there might need to be some alteration of the text there (as the example can't get our code to be slower than without atomslice). Maybe try adding a map for a single frame?

I probably should have actually read the text instead of just removing every bit of code that said `ContactMap`.... anyway, I think what's there now illustrates the original point.

PR #87, opened 2020-09-03
IndexManagers for atom slicing

dwhswenson:
As I [mentioned when atom slicing was first introduced](https://github.com/dwhswenson/contact_map/pull/35#pullrequestreview-163298417), the extra code for atom slicing showed that there's a separation of concerns issue in the design -- we really need a separate class to manage indexing. In fact, I think the even better thing is to move all the caching of indices into a separate class just for that. This means also moving the `residue_ignore_atom_idxs` and `residue_query_atom_idxs` out of the main `ContactObject` class, too.

Overall, this PR removes a lot of things from the `ContactObject` API. However, all of these things will still be available under `contact_object.indexer` (with a few minor changes to usage):

Changes to the public API (as documented in, e.g., [the `ContactFrequency` docs](https://contact-map.readthedocs.io/en/latest/api/generated/contact_map.ContactFrequency.html))

* `idx_to_s_idx()`: Now a dict called `indexer.sliced_idx`
* `s_idx_to_idx()`: Now a dict called `indexer.real_idx`
* `convert_atom_contacts()` => `indexer.convert_atom_contacts()`
* `slice_trajectory()` => `indexer.slice_trajectory()`
* `residue_query_atom_idxs` => `indexer.residue_query_atom_idxs`
* `residue_ignore_atom_idxs` => `_residue_ignore_atom_idxs`  (in the `ContactObject`, but private API)

These aren't really user-facing in the same way that, e.g., `atom_contacts` is. They're more about the internals (although they can be useful for sanity checks). I think this makes a more coherent and readable API. Similarly, I think it would be good to move `contact_map` private, but that's for another PR.

This does have some effect on storage: currently we store and reload some of these dicts. I think that can just be removed without problem, though. Everything in the indexer can be derived from things we will still store (`query`, `haystack`, and `all_atoms`).

### Performance

I wanted to ensure that there wasn't any significant performance issue introduced by this approach. Here's a little script I wrote, based on the atom slice notebook:

<details><summary>performance.py</summary>

```python
import timeit
import cProfile

import mdtraj as md
from contact_map import ContactFrequency, ContactTrajectory, version

traj = md.load("5550217/kras.xtc", top="5550217/kras.pdb")[:50]
topology = traj.topology

atoms = list(range(topology.n_atoms))
used_atoms = atoms[1:]
heavies = topology.select("protein and element != H")

print(f"     Current commit: {version.full_version}")
n_tries = 10
freq_test = "ContactFrequency(traj[0], query=used_atoms, haystack=used_atoms)"
traj_test = "ContactTrajectory(traj, query=heavies, haystack=heavies)"


ContactFrequency._class_use_atom_slice = True
timing = timeit.Timer(freq_test, globals=globals())
times = timing.repeat(number=n_tries)
t_sliced = min(times)
print(f" Freq: Atom slice enabled: {t_sliced / n_tries}")
cProfile.run(freq_test, 'with_atomslice.pstats')

ContactFrequency._class_use_atom_slice = False
timing = timeit.Timer(freq_test, globals=globals())
times = timing.repeat(number=n_tries)
t_unsliced = min(times)
print(f"Freq: Atom slice disabled: {t_unsliced / n_tries}")
cProfile.run(freq_test, 'no_atomslice.pstats')

ContactTrajectory._class_use_atom_slice = True
timing = timeit.Timer(traj_test, globals=globals())
times = timing.repeat(number=5)
t_unsliced = min(times)
print(f" Traj: Atom slice enabled: {t_unsliced / 5}")
cProfile.run(traj_test, 'traj.pstats')
```

</details>

And output from a few commits:

```
     Current commit: 0.6.0+g2c8d437  # release 0.6.0
 Freq: Atom slice enabled: 0.15540173510000005
Freq: Atom slice disabled: 0.12252338769999974
 Traj: Atom slice enabled: 1.7406494088000002
```

```
     Current commit: 0.6.1.dev0+g0f5201b  # this PR
 Freq: Atom slice enabled: 0.1577422553
Freq: Atom slice disabled: 0.12046521950000014
 Traj: Atom slice enabled: 1.675931647400001
```

For `ContactFrequency`, with or without atom slicing, that timing difference is within the errorbars. There's a small improvement in `ContactTrajectory` (although it was a more significant problem in a Real Research™ project of mine, where this PR also massively reduces memory usage). This is because I added the ability to provide the indexer to the `from_contacts` constructor, which prevents it from being recreated for each frame. The overhead cost for atom slicing is entirely in the `convert_contacts` function, which is only called once per `ContactFrequency` (and I don't see any way to make it significantly faster).


dwhswenson:
Passes tests and maintains coverage. Ready for review.

contact_map/atom_indexer.py
```diff
@@ -0,0 +1,107 @@
...
+                   for pair, value in atom_contacts.items()}
+        return collections.Counter(result)
+
+    def slice_trajectory(self, trajectory):
+        # Prevent (memory) expensive atom slicing if not needed.
+        # This check is also needed here because ContactFrequency slices the
+        # whole trajectory before calling this function.
+        if len(self.all_atoms) < trajectory.topology.n_atoms:
+            sliced = _atom_slice(trajectory, self.all_atoms)
+            sliced_trajectory = _atom_slice(trajectory, self.all_atoms)
```
sroet, line 76: excess defined variable?
```suggestion

```

contact_map/atom_indexer.py
```diff
@@ -0,0 +1,107 @@
...
+        return sliced
+
+
+class IdentityIndexer(object):
+    """Indexer when not using atom slicing.
+    """
+    def __init__(self, topology, real_query, real_haystack, all_atoms):
+        self.all_atoms = all_atoms
+        self.topology = topology
+        identity_mapping = {a: a for a in range(topology.n_atoms)}
```
sroet, line 88: As I would classify this as "performant" code, maybe only make dicts for the atoms used to reduce Memory (which might be less than all atoms, even without atom_slice)?
```suggestion
        identity_mapping = {a: a for a in all_atoms)}
```

feel free to ignore

contact_map/tests/test_contact_map.py
```diff
@@ -84,7 +84,7 @@ def _contact_object_compare(m, m2):
     assert m.query == m2.query
     assert m.haystack == m2.haystack
     assert m.n_neighbors_ignored == m2.n_neighbors_ignored
-    assert m._atom_idx_to_residue_idx == m2._atom_idx_to_residue_idx
+    # assert m._atom_idx_to_residue_idx == m2._atom_idx_to_residue_idx
```
sroet, line 87: This can be dropped instead of commented
```suggestion

```

contact_map/tests/test_contact_map.py
```diff
@@ -186,8 +185,8 @@ def test_to_dict(self, idx):
         assert dct['haystack'] == list(range(10))
         assert dct['all_atoms'] == tuple(range(10))
         assert dct['n_neighbors_ignored'] == 0
-        assert dct['atom_idx_to_residue_idx'] == {i: i // 2
-                                                  for i in range(10)}
+        # assert dct['atom_idx_to_residue_idx'] == {i: i // 2
+                                                  # for i in range(10)}
```
sroet, line 189: can be dropped
```suggestion

```

contact_map/tests/test_contact_map.py
```diff
@@ -324,7 +323,8 @@ def test_atom_slice(self, idx, use_atom_slice):
 
         # Test sliced indices
         sliced_idx = [0, 1, 2, 3]
-        real_idx = [map0b.s_idx_to_idx(i) for i in sliced_idx]
+        # real_idx = [map0b.s_idx_to_idx(i) for i in sliced_idx]
```
sroet, line 326: can be dropped
```suggestion

```

sroet (changes requested):
Some leftover commented out code in the tests and an extra `atom_slice` call.

sroet:
Another memory reducing step that might be possible later on (which is implemented in the current version of the code):
We know sliced indices are continuous from `0` to `len(all_atoms)-1` so you don't have to save the dict mapping both ways.

You can instead have a `a = tuple(all_atoms)` where using `a[sliced_idx]` gives you the real one and `a.index(real_idx)` gives you the sliced one.

This should be done in a different PR though, and is no reason to block this one

contact_map/atom_indexer.py
dwhswenson: Nice catch; that's certainly just a waste of time!

contact_map/atom_indexer.py
dwhswenson: I agree, and implemented it this way to start. However, that broke tests here (`KeyError: 0` on line 327):

https://github.com/dwhswenson/contact_map/blob/0f5201bd80be33b0207d687b9bc884e11e4a72e3/contact_map/tests/test_contact_map.py#L324-L331

I tried to change the tests as little as possible to avoid breaking intended behavior. The change to that test that works with the proposed change is:

```python
        # Test sliced indices
        sliced_idx = [0, 1, 2, 3]
        real_idx = [map0b.indexer.real_idx.get(i, None) for i in sliced_idx]
        if map0b._use_atom_slice:
            assert real_idx == [1, 4, 5, 6]
        else:
            assert real_idx == [None, 1, None, None]
```

I find it a little counterintuitive that `indexer.real_idx[0]` doesn't exist.

Can go either way; leaning toward leaving as is because this isn't *that* performance-focused. The only use case where this is a likely performance hit is where you're doing all-atom contacts for many different systems; e.g., proteomics. (Not all atoms? Use slicing to speed you up. Same structure? Use `ContactTrajectory`; this only happens once.)

sroet (approved):
LGTM, feel free to merge

PR #91, opened 2020-10-21
API cleanup: private `ContactObject._contact_map`; remove `frames` from `ContactFrequency`

dwhswenson:
This PR includes some additional API changes that I want to make before the next release (a lot of 0.7 will be API changes).

Here are the things that this breaks:

- [x] `ContactObject.contact_map` => `ContactObject._contact_map`: Taking it private because it doesn't make sense as a user-facing method.
- [x] Remove `ContactFrequency.frames`, as well as the `frames` parameter in `ContactFrequency.__init__`. This can instead by accomplished by slicing the MDTraj trajectory.

There is an argument that the `frames` parameter allowed you to create different multiple contact analyses from a single MDTraj trajectory, without needing to double memory usage. But I think it overcomplicated the main `__init__` call, which already has too many parameters.


sroet (approved):
LGTM, feel free to ignore CC

PR #92, opened 2020-10-23
Enable binder

sroet:
This will try to setup the right files for binder to be able to run our examples.

This will mainly use:
- `environment.yml` to build the `example` dependencies
- Maybe use a `postBuild` to pull in the example data.

Another PR will actually add the binder links to out repo.

current binder link for this branch:
[Binder](https://mybinder.org/v2/gh/sroet/contact_map/try_binder?filepath=%2Fexamples)

Working examples (will update as I go):

- [X] Comparing different data structures
- [X] concurrences 
- [X] contact_map 
- [x] contact_map_without_atom_slice
- [X] contact_trajectory
- [x] dask_contact_frequency

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -4,6 +4,9 @@ dependencies:
   - python
   - pip
   - matplotlib
+  - nglview
+  - dask
+  - distributed
   - pip:
     - tables
     - .
--- a/environment.yml
+++ b/environment.yml
@@ -6,3 +6,4 @@ dependencies:
   - matplotlib
   - pip:
     - tables
+    - .
--- a/environment.yml
+++ b/environment.yml
@@ -4,3 +4,5 @@ dependencies:
   - python
   - pip
   - matplotlib
+  - pip:
+    - tables
--- a/environment.yml
+++ b/environment.yml
@@ -1,0 +1,6 @@
+channels:
+  - conda-forge
+dependencies:
+  - python
+  - pip
+  - matplotlib
--- a/postBuild
+++ b/postBuild
@@ -7,3 +7,4 @@ cd examples
 mkdir $figshare_ID
 curl -Lk -o ${figshare_ID}.zip https://ndownloader.figshare.com/articles/5550217?private_link=453b1b215cf2f9270769
 unzip -d $figshare_ID ${figshare_ID}.zip
+rm $figshare_ID.zip
--- a/postBuild
+++ b/postBuild
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+figshare_ID=5550217
+
+# Dowload the extra example files
+cd examples
+mkdir $figshare_ID
+curl -Lk -o ${figshare_ID}.zip https://ndownloader.figshare.com/articles/5550217?private_link=453b1b215cf2f9270769
+unzip -d $figshare_ID ${figshare_ID}.zip


dwhswenson:
Definitely +1 on the idea. For the postBuild, the [curl-based figshare downloader](https://gitlab.e-cam2020.eu/dwhswenson/ops_additional_examples/-/blob/master/devtools/figshare_dl_extract.sh) I used in the `ops_additional_examples` repo might be useful.

sroet:
> Definitely +1 on the idea. For the postBuild, the curl-based figshare downloader I used in the ops_additional_examples repo might be useful.

I had to hard-code it in the script, as we use a slightly different link for the example figshare :wink: 

Another thing I ran into; the servers we run this on are pretty heterogeneous (depending on the hosting we get), the first time I ran through them, the "traj" contact frequencies took 3,5 min while the current run they take 30s. 
We might need to warn for this in the documentation

dwhswenson:
> Another thing I ran into; the servers we run this on are pretty heterogeneous (depending on the hosting we get), the first time I ran through them, the "traj" contact frequencies took 3,5 min while the current run they take 30s.
> We might need to warn for this in the documentation

Yeah, I gave a quick try out of curiosity before I commented, and contact traj was taking forever. Agree on comment in the example, but it should be brief.

BTW, this just works with the whole `examples/` directory, right? I've started on what's basically a complete rewrite of the main example notebook (splitting it into 3), so let's avoid anything where we'd create conflicts. Although I just have 3 new notebooks to replace the existing one.

sroet:
> BTW, this just works with the whole examples/ directory, right? I've started on what's basically a complete rewrite of the main example notebook (splitting it into 3), so let's avoid anything where we'd create conflicts. Although I just have 3 new notebooks to replace the existing one.

This will only add files in the main directory (as that is the only place `mybinder.org` looks for the files to build the dockers. 
It is just to get all the dependencies and data in place so that we could start linking to binder. 

A later PR will start adding badges to the docs. (which is easier to check if `master` is in shape to build a `binder` from)

sroet:
on that note, this should now be ready for a review

dwhswenson:
> I had to hard-code it in the script, as we use a slightly different link for the example figshare 😉

Any reason not to make that figshare public now? I mean, it's already relatively public since it is used as an example here 😂
-- plus the ArXiV paper is up and about to start collecting citations. Maybe we need to check with @jvreede, but I assume there's no problem making that trajectory officially public now.

dwhswenson (approved):
Code changes are obviously fine.... I'm still waiting to run it myself (taking a very long time to build), but then I'll be happy to merge it!

sroet:
> Maybe we need to check with @jvreede, but I assume there's no problem making that trajectory officially public now.

+1 on checking in with her

sroet:
>  I'm still waiting to run it myself (taking a very long time to build), but then I'll be happy to merge it!

Feel free to merge whenever you feel you tested it enough. (The builds should only happen once per commit, so that should normally not be an issue)

dwhswenson:
All looks good to me!

I think the place to mention performance issues will be next to the badge to launch it. It's binder-specific, and that way we don't distract in the main example text. I think the place for that text (and the button) is right after the link to the docs in the readme (before the mini-installation explanation). Something like:

---

Full documentation is at http://contact-map.readthedocs.io/.

Try it out online: [binder badge] (Note: the performance of the online servers can vary widely.)

---

> > Maybe we need to check with @jvreede, but I assume there's no problem making that trajectory officially public now.
> 
> +1 on checking in with her

Just sent her a text with a link here, in case she doesn't get the email notifications.

dwhswenson:
@sroet : Jocelyne replied by text ok'ing the making that figshare public.

sroet:
@sroet : Jocelyne replied by text ok'ing the making that figshare public.

slight hickup with that one. It is linked against my (dead) uva account... I can make a public copy of it?

dwhswenson:
> slight hickup with that one. It is linked against my (dead) uva account... I can make a public copy of it?

Ugh I hate those stupid policies of killing emails... Sure, a copy works. No problem changing the link here (and the old link is still valid).

PR #93, opened 2020-10-23
Add binder links

sroet:
after pr #92 
This PR adds the repective binder links to the 
- [X] `README.md`
- [x] examples page of the documentation
- [X] add a link to binder for every notebook
- [X] enable `jinja2` rendering for all `.rst` your docs

(@dwhswenson any place missing? for example I can add a link to each `ipynb` that would start binder for themself)

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -64,6 +64,25 @@ extensions = [
     'nbsphinx',
     'IPython.sphinxext.ipython_console_highlighting'
 ]
+# Adding jinja passthrough for html
+# Taken from https://www.ericholscher.com/blog/2016/jul/25/integrating-jinja-rst-sphinx/
+html_context = {"release": release,
+                "version": version}
+def rstjinja(app, docname, source):
+    """
+    Render our pages as a jinja template for fancy templating goodness.
+    """
+    # Make sure we're outputting HTML
+    if app.builder.format != 'html':
+        return
+    src = source[0]
+    rendered = app.builder.templates.render_string(
+        src, app.config.html_context
+    )
+    source[0] = rendered
+
+def setup(app):
+    app.connect("source-read", rstjinja)
 
 # Napolean settings
 napoleon_google_docstring = False
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -385,3 +385,27 @@ texinfo_documents = [
 
 # Example configuration for intersphinx: refer to the Python standard library.
 intersphinx_mapping = {'https://docs.python.org/': None}
+
+# Try autogenerating binder links, adapted from
+# https://github.com/spatialaudio/nbsphinx/blob/345f406e6f98d584fa493c788099e68b218f3a30/doc/conf.py#L54
+nbsphinx_prolog = r"""
+{% set docname = env.doc2path(env.docname, base='').replace("/nb", "") %}
+{% if "dev" in env.config.release %}
+{% set version = "master" %}
+{% else %}
+{% set version = "v"+env.config.version %}
+{% endif %}
+.. only:: html
+
+    .. role:: raw-html(raw)
+        :format: html
+
+    .. nbinfo::
+
+        This page was generated from `{{ docname }}`__.
+        Interactive online version:
+        :raw-html:`<a href="https://mybinder.org/v2/gh/dwhswenson/contact_map/{{ version }}?filepath={{ docname }}"><img alt="Binder badge" src="https://mybinder.org/badge_logo.svg" style="vertical-align:text-bottom"></a>`
+
+    __ https://github.com/dwhswenson/contact_map/blob/
+        {{ version  }}/{{ docname }}
+"""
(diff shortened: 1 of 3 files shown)


dwhswenson:
I think README and examples should be enough. BTW, part of my upcoming examples refresh involves adding numbers to the notebooks. That will:

1. Order them correctly in the docs (although we could do that manually)

2. Order them correctly in the Binder listing.

sroet:
> I think README and examples should be enough. 

Perfect, then this one is ready for a review

> BTW, part of my upcoming examples refresh involves adding numbers to the notebooks. That will:
> 
>     1. Order them correctly in the docs (although we could do that manually)
> 
>     2. Order them correctly in the Binder listing.

Sounds great :smile:

sroet:
@dwhswenson (just FYI for the possibilities of binder; the strangest `binder` setup I have done allows you to run Coupled-Cluster calculations in your browser [Binder](https://mybinder.org/v2/gl/sroet%2Fet_tutorial/master))

docs/examples/index.rst
```diff
@@ -4,11 +4,19 @@ Examples
...
     :glob:
 
     nb/*
 
+You can also try them out online directly from your browser: 
+
+.. image:: https://mybinder.org/badge_logo.svg
+ :target: https://mybinder.org/v2/gh/dwhswenson/contact_map/master?filepath=%2Fexamples
+
+(Note: the performance of the online servers can vary widely.)
```
dwhswenson, line 20: Go ahead and move this above the toctree -- as the contents grow, I think it'll be better to have this above the fold. Also, probably better to use an inline image, which can be accomplished in RST [using a substitution definition](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#substitution-definitions).

dwhswenson (changes requested):
Suggesting a little modification to move the Binder link higher up in the main examples page. Also, here's a better discussion of images/links/substitutions in RST than the one I linked in that comment:

* https://stackoverflow.com/a/14387603

Not necessarily for this PR, but nbsphinx has a trick for adding frontmatter that can automatically include a link to the relevant notebook on Binder: 

* https://nbsphinx.readthedocs.io/en/0.4.1/prolog-and-epilog.html

(That link is also an example of this specific use in action.)

Consider that optional; if it is a pain to do, don't worry, but if it's easy, it would be nice. I don't mind having a Binder link for each page on RTD; I just didn't want to put Binder links in each ipynb, since it doesn't make sense if you've already loaded it locally.

docs/examples/index.rst
sroet: done

sroet:
> Consider that optional; if it is a pain to do, don't worry, but if it's easy, it would be nice. I don't mind having a Binder link for each page on RTD; I just didn't want to put Binder links in each ipynb, since it doesn't make sense if you've already loaded it locally.

So I got it working, but this only works for releases if the docs are build from a release install of `contact_map` (because it checks if the installed version of cantact map contains "dev")

sroet:
one thing that I could not figure out:
The link properly in `docs/examples/index.rst` can not be set depending on the version as that one does not seem to pass through jinja2 before converting to html

sroet:
Alright, I also added `jinja2`  rendering for all the `.rst`, so you can do some logic in them as well. (All the `jinja2` code in the `.rst` has access to all variables in the `html_context` dictionary.)

docs/conf.py
```diff
@@ -64,6 +64,25 @@
     'nbsphinx',
     'IPython.sphinxext.ipython_console_highlighting'
 ]
+# Adding jinja passthrough for html
+# Taken from https://www.ericholscher.com/blog/2016/jul/25/integrating-jinja-rst-sphinx/
+html_context = {"release": release,
+                "version": version}
```
sroet, line 70: These will be the variables that are available to `jinja2` code in the `.rst`

docs/conf.py
```diff
@@ -64,6 +64,25 @@
...
+def rstjinja(app, docname, source):
+    """
+    Render our pages as a jinja template for fancy templating goodness.
+    """
+    # Make sure we're outputting HTML
+    if app.builder.format != 'html':
+        return
+    src = source[0]
+    rendered = app.builder.templates.render_string(
+        src, app.config.html_context
```
sroet, line 80: this `html_context` is the dictionary build on line `69` and will provide the variables that `jinja2` will put in scope during rendering.

docs/examples/index.rst
```diff
@@ -1,14 +1,28 @@
...
 We have several examples to illustrate various features of the code. These
 notebooks have been rendered here for the web, but the originals are
 found in the ``examples/`` directory of the package, and you can run them
-yourself!
+yourself! 
+
+You can also try them out online directly from your browser: |binder|_
+(Note: the performance of the online servers can vary widely.)
+
+|binder|_ binder link
```
dwhswenson, line 19:
```suggestion
```

dwhswenson: This shows the badge and "binder link", which isn't what you wanted

dwhswenson (changes requested):
One very small change, then it looks good!

sroet:
Thanks! Corrected that one

dwhswenson (approved):

PR #90, opened 2020-10-01
ContactCount.to_networkx

dwhswenson:
A sparse matrix (such as a contact map) can be considered equivalent to a network. This PR adds a direct function to make a NetworkX graph from a `ContactCount`.

This constructor allows you to also add our graph to an existing graph by specifying the `graph` keyword. Additionally, you can select whether you want the graph to use the indices as nodes or the topology objects. 

Another way to create a NetworkX graph from a `ContactCount` would be:

```python
g = nx.from_scipy_sparse(contact_count.sparse_matrix)
```

There are options possible with `nx.from_scipy_sparse` that are not possible with `ContactCount.to_networkx`. See [NetworkX docs on it](https://networkx.github.io/documentation/stable/reference/generated/networkx.convert_matrix.from_scipy_sparse_matrix.html).

PR #94, opened 2020-10-24
[non-public API-break] Autozoom plots

### LINKED ISSUE #89: "Autozoom" plotting
I find that I'm frequently doing something like:

```python
plt.xlim(*contacts.query_residue_range)
plt.ylim(*contacts.haystack_residue_range)
```

This zooms my plot into just the query-by-haystack region. This should be the default behavior.

To get there, I'm suggesting the following:

1. [API BREAK] Replacing `n_x`, `n_y` in `ContactCounter` by `query_range` and `haystack_range`. The original idea of separate `n_x` and `n_y` was to do something like this, but the original implementation didn't work, and now everything just uses `n_x = n_y = n_total` (where `total` can be `residues` or `atoms`).
2. Adding a default parameter `autozoom=True` to the `.plot` and `.plot_axes` methods.
3. Updating the dpi warning in `.plot` to know what the autozoomed size will be.
4. [API BREAK] Remove `ContactObject.query_residue_range`, `ContactObject.haystack_residue_range`. They exist entirely for this purpose.

For discussion: is the query the x axis or the y axis?

sroet:
Should close #89

**[non-public API-break]** This code breaks some non-public API, see **Other Changes** below

Description of the new behaviour:
For all `ContactMaps`, except `ContactDifference` (and its subclasses):
It will `contacts.plot()` will plot `query`/`query_residues` on the `x`-axis and `haystack`/`haystack_residues` on the `y`-axis.

For `ContactDifference` the `query` and `haystack` and the `_residue` equivalents might not make sense so instead they use the new default, which tries to plot all points with as little amount of "double" (due to symmetry) points as it can.

## The new n_x, n_y defaults
We construct the lower triangle of our data by sorting the keys (indices) and then we look at `x` and `y` dimensions of the smallest rectangle required to cover all the datapoints in this triangle. 
Effectively trying to see if there is a set of `low` and a set of `high` indices that can be used to describe all data points (like a `query` and a `haystack`). We then take the shortest of the 2 dimensions as `n_x` and the longest as `n_y`.

## Other changes
- The `sparse_matrix` and `pandas.DataFrame` now are guaranteed to be square,  `max_size` by `max_size` (being effectively equal to the old implementation for all classes.
- The `plot` warning now takes into acount the actual range of data being plotted
- `haystack_residues_indices` and `query_residues_indices` are now calculated and stored at the initialization.
- **[non-public API-break]** With the above point `_range_from_object_list()` was not needed anymore and removed.
- Added a `query_range` and `haystack_range` like  their `*_residue_range` equivalents.

## Plotting everything again
The following code can be used to plot all points of the contact map:
```
fig, ax = residue_contacts.plot()
ax.set_xlim(*residue_contacts.total_range)
ax.set_ylim(*residue_contacts.total_range)
```

### Some other info that was here before the last edit:

Things I like some feedback on (preferably from @dwhswenson ):
This was offered as a possibility for the user:
```
fig, ax = residue_contacts.plot()
ax.set_xlim(0, residue_contacts.max_size)
ax.set_ylim(0, residue_contacts.max_size)
```

~What should `total_range` be? 
The total range we have data (min/max on the `counter`)?
 or `0` to `topology.n_atoms`/`topology.n_residues` (this would require some more code as topology is not something the ContactCounts  have)?~

Nvm, we need the size anyway for casting to matrices in a sensible way, so I added it as a keyword.

~Contains similar code as #95, but that can be split out if this is ready to be merged before that one is in~

Current issues to solve:
- [x] this does not work (yet) for contact differences
- [x] examples need updating
- [x] tests don't (yet) pass
- [x] add test for the new code


sroet:
Alright @dwhswenson this one is ready for a review, please let me know if this is what you intended by #89

sroet:
@dwhswenson A friendly ping to make sure this is still on your radar :wink:

sroet:
@dwhswenson , friendly monthly ping to make sure it is still on a list somewhere :)

sroet:
@dwhswenson Just FYI; this PR needs to be squashed before merging, as it contains multiple commits altering all the notebooks, which will lead to repository bloat.

sroet:
Also, I still don't agree with the codeclimate failure

dwhswenson:
@sroet : Could you resolve conflicts on this branch? I'm hoping to try reviewing this soon. Since this has a lot of notebook changes, I want to hold off on notebook updates in #90 and #110 until this is in, to limit conflicts.

sroet:
> @sroet : Could you resolve conflicts on this branch? I'm hoping to try reviewing this soon. Since this has a lot of notebook changes, I want to hold off on notebook updates in #90 and #110 until this is in, to limit conflicts.

All done :)

dwhswenson:
> Also, I still don't agree with the codeclimate failure

Ah, now I remember these. (Old link had stopped working) Yes, I disagree as well. It's clear that they add "mass" to their code similarity measure when you use kwargs. That is nonsense. When these first popped up, I actually started going into their codebase to try to figure out how I might fix it, but between the fact that most of the code is CodeClimate extensions to a 3rd party code, plus the fact that it is mostly written in Ruby, I gave up.

dwhswenson:
Before I get into the code, a question about the overall approach:

If I understand correctly, the x vs y axis can be either query or haystack depending on the number of residues. I think that may cause confusion for the users. Isn't it easier to just say that default x-axis is the query and default y-axis is the haystack? This seems like it would be more consistent (and, in any case, give identical results for users who follow our recommendation to keep the query smaller than the haystack.)

sroet:
> If I understand correctly, the x vs y axis can be either query or haystack depending on the number of residues. I think that may cause confusion for the users. Isn't it easier to just say that default x-axis is the query and default y-axis is the haystack? This seems like it would be more consistent (and, in any case, give identical results for users who follow our recommendation to keep the query smaller than the haystack.)

Not really, that is just the `ContactCount` default (which has no idea of the existence of queries or haystacks). 

For all classes that use `ContactDict.atom_contacts` (or `residue_contacts`) we explicitly set `n_x` to `query_range` and `n_y` to `haystack_range`. 

If we make a `ContactCount` from somewhere else (like in `ContactDifference`), then we use the data inside the counter to get the best guess (these might not even be complete `queries` or `haystacks` if there are certain contacts missing in the `counter`s)

contact_map/contact_count.py
```diff
@@ -57,6 +57,49 @@ def _patch_from_spmatrix(cls, data):  # -no-cov-
...
+
+def _get_low_high_counter_range(counter):
+    """Give the (min, max + 1) for both the low and high keys in counter"""
+    keys = [tuple(sorted(list(i))) for i in counter.keys()]
+    if len(keys) == 0:
+        return (0, 0), (0, 0)
+
+    lows = sorted([i for i, j in keys])
+    highs = sorted([j for i, j in keys])
+    return (lows[0], lows[-1]+1), (highs[0], highs[-1]+1)
```
dwhswenson, line 83: Use `min` and `max` instead of `sorted` and getitem `[0]`, `[-1]` here. (Not that I think this will be a performance bottleneck, but we might as well do it right! Also, I think it easier to read.)

I think you can get `lows, highs = zip(*keys)`, too.

contact_map/contact_count.py
```diff
@@ -57,6 +57,49 @@ def _patch_from_spmatrix(cls, data):  # -no-cov-
...
+
+
+def _get_sorted_counter_range(counter):
+    """Return smallest range, longest range for the low and high counter"""
+    low, high = _get_low_high_counter_range(counter)
+    if low[1]-low[0] > high[-1]-high[0]:
+        return high, low
+    return low, high
```
dwhswenson, line 91: Readability here would be improved with explicit `else` statement and single return. That is, something like:

```python
if ...:
    n_x, n_y = high, low
else:
    n_x, n_y = low, high
return n_x, n_y
```

contact_map/contact_count.py
```diff
@@ -83,16 +126,31 @@ class ContactCount(object):
...
-    def __init__(self, counter, object_f, n_x, n_y):
+    def __init__(self, counter, object_f, n_x=None, n_y=None, max_size=None):
         self._counter = counter
         self._object_f = object_f
+        self.total_range = _get_total_counter_range(counter)
+        n_x, n_y = _sanitize_n_x_n_y(n_x, n_y, counter)
         self.n_x = n_x
         self.n_y = n_y
+        self.n_x_min, self.n_x_max = _int_or_range_to_tuple(n_x)
+        self.n_y_min, self.n_y_max = _int_or_range_to_tuple(n_y)
```
dwhswenson, line 147: It seems to me that there's enough (repeated) logic here that maybe it's worth refactoring into some small class `ContactPlotRange` (or any better name you can think of) so that here it could just be `self.n_x = ContactPlotRange(n_x, counter)`, and `self.n_x.min` and `self.n_x.max` would give the desired info (and perhaps `self.n_x.as_input`, although I'm not sure whether that is necessary to preserve here). You could also then create `@property` something like:

```python
@property
def range_length(self):
    return self.max - self.min
```

This gets used in, e.g., the DPI check, and would probably make that a little easier to read.

dwhswenson (changes requested):
Main things:

1. Please add something brief to the `changing_defaults.ipynb` notebook, in the section on changing query/haystack, to show how to use the full range (something like what you have in the main post here with `residue_contacts.max_size`). You can point to further information the the notebook on customizing plot appearance, but I think we need a brief mention of how to zoom out there, since we now default to zooming in.
2. I think it might be worth creating a class that `n_x` and `n_y` are each an instance of. I thought of the name `ContactPlotRange`, but other names might be better. The idea is that a lot of stuff is getting reused for both x and y, including adding more to the `ContactObject` namespace (`n_x_min`, `n_y_min`) and using logic (`n_x_max - n_x_min`) that gets repeated for both x and y in a way that makes me think it's worth a refactor.


Other than that, a couple small suggestions that I think will improve readability.

sroet:
@dwhswenson Did the refactor (moved relevant code to `plot_utils`), and added the plot to the "changing the defaults" ipynb please have another look

dwhswenson (commented):
Thought: Should `ContactPlotRange` be public API? I lean toward no. Functionality like `contacts.n_x.range_length` is public API, but the specifics of what `n_x` is such that this is available is not. So I tend to think `ContactPlotRange` => `_ContactPlotRange`. (But if you disagree, I can definitely be convinced otherwise.)

sroet:
> Thought: Should ContactPlotRange be public API? 

I have no preference on this, so I will make it private.

contact_map/tests/test_plot_utils.py
```diff
@@ -6,7 +6,7 @@
 from .utils import *
 
 from contact_map.plot_utils import *
-
+from contact_map.plot_utils import _ContactPlotRange
```
sroet, line 9: This is needed as star imports don't import private classes and variables

dwhswenson: Yeah, this is the reason I didn't make it a batch of suggestions. I couldn't make a suggestion on the code that hadn't changed. 😉

dwhswenson (approved):

dwhswenson:
LGTM. Squash-merging as suggested, given many changes to notebooks (almost forgot to do that!)

PR #97, opened 2020-10-28
Fix spaces in links

sroet:
I looked at the latest release
The good news: It builds the links against the `v0.7.0` tag.

The bad news: I apparently did not check all the links properly enough and the one with spaces (`Comparing different structures`) led to dead links. (Binder reverts to the example directory, but the github link just straight up fails.)

This fixes it by replacing the spaces with `%20` which do work in links.


dwhswenson (commented):
It's probably better to use [jinja's `urlencode` filter](https://jinja.palletsprojects.com/en/2.11.x/templates/#urlencode). This will also catch other URL issues in the future.

sroet:
> It's probably better to use jinja's urlencode filter. This will also catch other URL issues in the future.

Thanks for the pointer, that seems like a better solution indeed

dwhswenson (approved):

Issue #99, opened 2020-10-30
Move from travis.org -> travis.com/github actions


sroet:
I just saw [this issue](https://github.com/dask/dask/issues/6781) pop up on some other code I was watching.

We should also probably think about migrating away (and auto-release as well)
Here is the travis-ci documentation on this issue: https://docs.travis-ci.com/user/migrate/open-source-repository-migration.

I am pretty sure this has to be done by the owner of this repo, so @dwhswenson this one is assigned to you

dwhswenson:
Did this a while ago. 😊 Looking back, must have been before releasing 0.5.1. You can tell because PRs with travis-ci.com show up in the "Checks" tab; PRs with travis-ci.org do not. I find a few aspects of the .com interface annoying compared to the .org (specifically, you go to the GitHub check, instead of directly to the Travis build, requiring an extra click), but they've made it clear for a while that .org is deprecated.

All my active projects except OPS are now on travis-ci.com. OPS is still on .org because I don't have access to some of the secure environment variables, and don't want to risk losing the ability to update the OPS website.

dwhswenson:
> I don't have access to some of the secure environment variables

Clarification: Nowadays, the best way to set secure environment variables is via the Travis web interface. But it used to be that you'd set them using the Travis CLI, which encrypts in a way that is tied to the repo name, and I wasn't 100% sure if the Travis-CI endpoint also mattered in that hashing. Web-set secure vars are definitely migrated. I think secure vars in encrypted secure tags in the .yml are also migrated, but I didn't want to really test that with my largest project.

sroet:
> > I don't have access to some of the secure environment variables
> 
> Clarification: Nowadays, the best way to set secure environment variables is via the Travis web interface. But it used to be that you'd set them using the Travis CLI, which encrypts in a way that is tied to the repo name, and I wasn't 100% sure if the Travis-CI endpoint also mattered in that hashing. Web-set secure vars are definitely migrated. I think secure vars in encrypted secure tags in the .yml are also migrated, but I didn't want to really test that with my largest project.

Yeah I understand the tendency to not want to test that with that big of a project :) (Couldn't you email/contact the Travis people to see if it was used in the hash or not?)

dwhswenson:
Sure, but really I need to get the domain under my ownership (which I'm pretty sure John will be happy to give, since apparently someone lost the domain for omnia.md)

PR #100, opened 2020-10-30
Update Travis badge to .com instead of .org

sroet:
The reason I thought #99 was not completed yet, was due to this badge still linking to .org. This fixes that


dwhswenson (approved):
Looks good to me.

I'm guessing that the badge must have been forwarding to the migrated repo, because [the last builds on .org](https://travis-ci.org/github/dwhswenson/contact_map/builds) were all failing. But the old link for the badge was pointing to a page saying the repo had migrated, so definitely correct to update these!

PR #106, opened 2021-03-08
New example notebooks

dwhswenson:
Ages ago, I mentioned that I was working on new docs. As we've added functionality and come across other user questions, we've tended to put all of that into the main `contact_map.ipynb` notebook. I think it has gotten to the point where that example has too many sort of side comments and it detracts from the core point of introducing functionality.

This PR is a reorganization of a lot of our docs. Overall, I'm imagining something like:

* `overview`: Mainly covers what the original `contact_map.ipynb` covered (even back in its [gist](https://gist.github.com/dwhswenson/41104499495adaf443882e229cba75ef) days)
* `changing_defaults`: `query`, `haystack`, `cutoff`, `n_neighbors_ignored`
* `customizing_plots`: multiple plots in one figure, changing color map, other plotting methods for performance
* `integrations`: exporting to other formats, performance considerations via dask or atom slicing

Those notebooks should replace `contact_map.ipynb`, `contact_map_without_atom_slice.ipynb`, and `dask_contact_trajectory.ipynb`.

I'll add notebooks here as they are completed enough for individual review (avoid needing to squash a bunch of notebook commits). Alternatively, I could do 1 PR per notebook, although I think that would leave the docs in a weird state of having the new notebooks at the same time as the old ones.


dwhswenson:
Eh, what the heck. It wasn't too much work to get the rest of these done.

This PR isn't quite finished, but before I drop the old notebooks and add these to the sphinx docs, I think it's worth having @sroet look over this and see if you agree with the general direction for the examples/docs.

I think I'm willing to merge this with the one TODO still in the notebook on plotting. I feel like the rest of these are a big enough improvement over the existing notebooks that it is better to get this in than to wait for perfection.

sroet (approved):
> This PR isn't quite finished, but before I drop the old notebooks and add these to the sphinx docs, I think it's worth having @sroet look over this and see if you agree with the general direction for the examples/docs.

General direction looks great :D I do have a couple nitpick comments that I will do in a review as soon as this is ready, please ping me again when that is the case :) (mainly about naming consistency, cross referencing etc)

dwhswenson:
@sroet: Should be ready now. I did some minor cleanup in those notebooks, but mostly just removed the old ones and added explicit mention of each notebook in the `examples/index.rst`. 

I renamed the Examples section as "User Guide/Examples", since it is more of a user guide now. The URL is still for `examples`, though.

examples/overview.ipynb
```diff
@@ -0,0 +1,806 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 12,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "atom_pairs = list(map(list, [k for k in frame_contacts.atom_contacts.counter.keys()]))"
+   ]
+  },
```
sroet, line 298: Is this leftover debug code or just some code without description (this is quite the invocation)?
Please either remove or add some description
```suggestion
```

examples/overview.ipynb
```diff
@@ -0,0 +1,806 @@
...
+     },
+     "execution_count": 20,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "traj_contacts.most_common_atoms_for_contact([val81, asn116])[0][0][0].index"
+   ]
+  },
```
sroet, line 560: Is this leftover debug code, or was this to indicate how to get the atom index for asn116-cb?
```suggestion
```
or add some description about what is done here

examples/changing_defaults.ipynb
```diff
@@ -0,0 +1,501 @@
...
+    "print(len(default_selection))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Note that the general assumption is that the query is no larger than the haystack. if this isn't obeyed, you'll still get correct answers, but some algorithms may be less efficient, and visualizations have also been designed with this in mind."
+   ]
+  },
```
sroet, line 72: Is this still true? (It has been a while since I thought about this requirement in the code)

If this is still true:
```suggestion
  {
   "cell_type": "markdown",
   "metadata": {},
   "source": [
    "Note that the general assumption is that the query is no larger than the haystack. If this isn't obeyed, you'll still get correct answers, but some algorithms may be less efficient, and visualizations have also been designed with this in mind."
   ]
  },
```
else:
```suggestion
  {
   "cell_type": "markdown",
   "metadata": {},
   "source": [
    "Note that the general assumption is that the query is no larger than the haystack and visualizations have been designed with this in mind."
   ]
  },
```

examples/changing_defaults.ipynb
```diff
@@ -0,0 +1,501 @@
...
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Refresher: The default parameters\n",
+    "\n",
+    "\n",
+    "\n",
+    "* *Which atoms are involved in the contact?* The default value is non-hydrogen, non-water atoms.\n",
+    "* *What is the cutoff distance is?* The default value is 0.45 nm.\n",
```
sroet, line 433: double "is"
```suggestion
    "* *What is the cutoff distance?* The default value is 0.45 nm.\n",
```

examples/custom_plotting.ipynb
```diff
@@ -0,0 +1,377 @@
...
+   "metadata": {},
+   "source": [
+    "## Putting multiple plots in one figure"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "You may want to plot multiple maps on one figure. This can easily be done by using the `plot_axes` method, which acts just like `plot` except that it requires that you give it a Matplotlib `Axes` object to plot into."
```
sroet, line 47: isn't "in" more correct here?
```suggestion
    "You may want to plot multiple maps in one figure. This can easily be done by using the `plot_axes` method, which acts just like `plot` except that it requires that you give it a Matplotlib `Axes` object to plot into."
```

examples/integrations.ipynb
```diff
@@ -0,0 +1,842 @@
...
+    "traj_contacts = ContactFrequency(traj)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Exporting contact data\n",
+    "\n",
+    "Contact Map Explorer makes it easy for you to get your contact data into other formats, so that you can use the tools you're already familiar with to perform your analysis. Note that all of these report as integers, so what you have is the index of the object in the MDTraj topology,"
```
sroet, line 47: what does "these"  mean in this sentence?
is it the contacts?:
```suggestion
    "Contact Map Explorer makes it easy for you to get your contact data into other formats, so that you can use the tools you're already familiar with to perform your analysis. Note that all of the contacts report as a pair of integers, so what you have is the index of the object in the MDTraj topology,"
```
or the residues/atoms
```suggestion
    "Contact Map Explorer makes it easy for you to get your contact data into other formats, so that you can use the tools you're already familiar with to perform your analysis. Note that the residues and atoms of these report as integers, so what you have is the index of the object in the MDTraj topology,"
```

sroet (changes requested):
Looks very good overall, here are the promised nitpick comments ;)

examples/changing_defaults.ipynb
dwhswenson: "***some*** algorithms ***may*** be less efficient" -- I'm very intentionally hedging here. 😆 I'm not saying which algorithms, or even that any actually are less efficient. Really, this is as much to developers: if you have an algorithm that needs one list to be bigger, assume the haystack is bigger. We keep telling the users to act that way, and then devs can take advantage of it.

dwhswenson:
@sroet : Barring objection, I think I'm going to change the name (URL) of the overview notebook to `contact_map.ipynb`. In the docs, it will still list under the heading "Overview" but this keeps the URL the same. Reason is that we've got decent Google mojo on this already. Here are our acquisition numbers, by channel and landing page:

image

Search is going straight to that notebook for 1/3 of our website visits, and there are a few others that directly load it or load it from a link somewhere else (5th and 10th most common sources of users).

sroet (approved):
LGTM

sroet:
>  Barring objection

No objection for me, please just make sure the auto-binder link still works ;) If that is the *only* further change feel free to merge after that without another round of review.

PR #88, opened 2020-09-04
Remove `ContactMap` (allowing rename `ContactFrequency` => `ContactMap`)

dwhswenson:
This is the next step in #82. Removes `ContactMap` and adds `PendingDeprecationWarning` on `ContactFrequency`.

This also bumps the version to 0.7.0.dev0 (which should have been done as soon as we broke the API in #87; changed API means we're developing on the next 0.x).

PR #111, opened 2021-04-11
Fix lines missing from test coverage

dwhswenson:
Our test suite currently fails to cover 5 lines. This PR should fix that.

* One line was not covered because the `with_colorbar` option wasn't passed from `ContactCount.plot` to `ContactCount.plot_axes` in #79. This PR adds that functionality back in.
* Four lines were not covered because the implementation at `ContactObject.atom_contacts` and `ContactObject.residue_contacts` are now overridden in all subclasses. It used to be that `ContactMap` used these, but `ContactMap` was removed in #88. This PR makes those explicitly abstract by raising `NotImplementedError`.

### DIFF
--- a/contact_map/contact_count.py
+++ b/contact_map/contact_count.py
@@ -204,7 +204,8 @@ class ContactCount(object):
 
         # Check the number of pixels of the figure
         self._check_number_of_pixels(fig)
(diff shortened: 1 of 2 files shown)


dwhswenson:
CodeCov agrees -- we are 💯!

Ready for review.

sroet (approved):
LGTM, feel free to merge

Issue #113, opened 2021-05-02
Renaming `master` branch to `main`


dwhswenson:
I'm planning to rename the `master` branch here to `main`. My thoughts on this are fully discussed in https://github.com/openpathsampling/openpathsampling/issues/956, so I won't repeat them here.

With the exceptions of Contact Map Explorer and OpenPathSampling, all projects that [I maintain on PyPI](https://pypi.org/user/dwhswenson/) have already made this switch -- I've been working up from the small projects to the larger ones.

[Here are instructions](https://stevenmortimer.com/5-steps-to-change-github-default-branch-from-master-to-main/) that users will need to follow to make this change (note you'll have to change the default branch **of your fork** via the GitHub web interface).

As for timing, I'd like to do this immediately after we release 0.8 (which is waiting on two currently open PRs plus one PR I need to open that renames `ContactFrequency` as `ContactMap`).

dwhswenson:
@sroet : When we do this, can you help ensure that the linked instructions are correct? I got a little confused when doing the openpathsampling-cli change -- the only one so far where I had to update a local fork instead of just the main repo. (I think I had just forgotten to make the change in my fork on GitHub, which was easy enough to fix.) I switched the OPS CLI sometime in early March or late Feb, and the only problem I've had since was that I had forgotten to fix the branch name in the GitHub Actions script.

sroet:
+1 on the rename, feel free to do it whenever you want.

Sure, I can run through the instructions once you decide to run it. 
I think I remembered that github gives you a button to change your fork if the upstream does this, and the only thing that is left is to pull main from your fork and delete the local master branch. But I might be misremembering things

PR #114, opened 2021-05-10
Update GA badge

sroet:
I noticed that the GA badge does not link to the actual action page, but just to the svg. This updates it to track the `master` branch and should be updated accordingly when #113 is done.


dwhswenson (approved):
LGTM. I noticed this the other day, but forgot to fix it. Thanks! (Also, GitHub has started including this in the default badge -- was pleased to see that when I added Actions to another project.)

PR #117, opened 2021-06-04
Remove genindex from docs

dwhswenson:
This removes the page at https://contact-map.readthedocs.io/en/latest/genindex.html from the docs. Prior to today, that page had been visited only once this year. It seems to me like it is more likely to confuse users than help them (it only indexes the API sections, and the search functionality/browsing the API docs should be easier to use than this index.)


sroet:
LGTM, that seems indeed like a confusing page that should already be done by using the search functionality

PR #118, opened 2021-07-26
Actually test for the Userwarning on custom cmaps

sroet:
I was looking into the reason why a `UserWarning` surfaced from our test suite.
```
=========================================== warnings summary ============================================
contact_map/tests/test_plot_utils.py::test_is_cmap_diverging[custom]
  /home/sander/github_files/contact_map/contact_map/plot_utils.py:31: UserWarning: Unknown colormap: Treating as sequential.
    warnings.warn("Unknown colormap: Treating as sequential.")

-- Docs: https://docs.pytest.org/en/stable/warnings.html
```

The reason was that the variable `cmap` was redefined before the line `if cmap == 'custom'` leading to the `if` statement never being true. 

This fixes that, makes sure we raise the correct warning and keeps the warning from showing up in the warning summary)
It also solves some whitespace complaints from my development environment.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -14,14 +14,11 @@ environment:
     PYTHONUNBUFFERED: 1
 
   matrix:
-    #- PYTHON: "C:\\Miniconda38-x64"
-      #CONDA_PY: "38" 
-      #ARCH: '64'
+    - PYTHON: "C:\\Miniconda38-x64"
+      CONDA_PY: "38" 
+      ARCH: '64'
     - PYTHON: "C:\\Miniconda37-x64"
       CONDA_PY: "37"
-      ARCH: '64'
-    - PYTHON: "C:\\Miniconda36-x64"
-      CONDA_PY: "36" 
       ARCH: '64'
 
 build: false
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -14,9 +14,6 @@ environment:
     PYTHONUNBUFFERED: 1
 
   matrix:
-    - PYTHON: "C:\\Miniconda38-x64"
-      CONDA_PY: "38" 
-      ARCH: '64'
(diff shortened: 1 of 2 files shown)


dwhswenson (commented):
Good catch. And this is why Ned Batchelder always says that you should include tests in your coverage reports -- we would have seen that part of the test wasn't being run.

(The main reasons I don't are (1) on OPS, the lower coverage numbers should encourage me to write more tests; (2) on all packages, I like using the the number of statements as a measure of code size, and feel like that should not include tests.)

There seems to be a problem with AppVeyor. I've re-run several times, getting the same error in the 3.6 build (3.7 passes). Our AppVeyor process is to use conda to install conda and conda-build, and then to build with conda-build. The failure is while installing conda-build, so it really can't be a result of this PR (but we don't do daily builds on AppVeyor, making it hard to judge when this broke).

Things I'll try in order to fix this:

1. Switch AppVeyor to testing with 3.7/3.8. They seem to have [a Miniconda with 3.8 available](https://www.appveyor.com/docs/windows-images-software/#miniconda), they didn't last I looked. (And still no 3.9.)
2. If that doesn't fix it (or while I'm trying that), @sroet, if you have a Windows machine available, could you try creating a conda env with Python 3.6 and then installing running `conda install -yq conda conda-build` from that env (to see if you can reproduce the error)?
3. Worst case, we can settle for only testing 3.7 on Windows. I'd actually rather test 3.7 and 3.9 (min/max of NEP29), but that requires more changes of our AppVeyor setup than I want to do right now.

dwhswenson (approved):
Dropped AppVeyor down to only build 3.7. Now this passes, and I'll merge. I think we'll move off of AppVeyor altogether; see #119.

PR #120, opened 2021-07-27
CI Cleanup: Windows builds on GitHub Actions; drop Python 3.6; better requirements handling

### LINKED ISSUE #119: Discussion: Moving off of AppVeyor
This issue is to discuss whether we should move away from AppVeyor as a testing platform. This is in part a response to problems in #118, although I've been thinking about this for a while.

Historically, we use AppVeyor because Travis-CI didn't have free Windows builds. Now we've moved from Travis-CI to GitHub Actions for our Linux builds. GitHub Actions does allow free Windows builds.

We also use two different build processes. The Linux builds install requirements with `pip`, getting packages from PyPI. The Windows builds use `conda-build`, getting packages from conda-forge. This can be nice because sometimes packages have slight differences in their dependencies (most frequently, they include some extras on conda-forge that don't come with the package when you install with `pip`). Testing both ensures that we aren't implicitly relying on something that only comes through one distribution mechanism.

The problems with the current Windows builds:

1. AppVeyor only allows one simultaneous build, so even running only 2 builds takes a while.
2. The actual build process (via conda-build) is slower on Windows, which exacerbates the problem of not having simultaneous builds.
3. With our current build process, we need to wait for AppVeyor to update Miniconda versions. Our build doesn't seem to work on their Miniconda38, and there is no Miniconda39. So we're limited to Python 3.7.
4. The conda-build based build is unnecessary. This is because I mimicked what MDTraj did. We don't actually need conda-build, since we're not compiling anything. We can install requirements with conda if we want. This will also get rid of the need for the recipe in devtools -- that recipe is completely independent of the conda forge feedstock recipe, and is only used for the AppVeyor build.

So I think it makes sense to move Windows builds over to GitHub Actions as well. I haven't used Windows builds on GHA yet, so I'd love to hear anyone else's experiences with it (@sroet, have you used that?)

A few specific points to consider:

* **Build matrix**: Which versions do we test against? I suggest only 2 (to keep build matrix smaller) and having those be the newest version of Python and the oldest NEP29-required version, and only testing against the release version of MDTraj (we test against development builds on Linux).
* **Where to get packages from**: I think we should drop conda-build, but maybe continue to get packages from conda.

Anything else to consider? Any thoughts on these?

dwhswenson:
Resolves #119.

This will drop AppVeyor and put Windows testing on GitHub Actions. I expect this to take a few steps. General approach I'm planning:

- [x] Refactor the testing workflow so that we have the following matrix variables (listed with their options): `MDTRAJ=['mdtraj-dev', 'mdtraj-release']`, `OS=['ubuntu', 'windows']`, `DEPS_FROM=['pip', 'conda']`, `CONDA_PY=['3.7', '3.8', '3.9']`
- [x] Add proper support for different installs based on each of the options.
- [x] Set up the matrix to run the following combinations:
  ```
  OS=ubuntu, DEPS_FROM=pip, MDTRAJ=mdtraj-release, CONDA_PY=3.7
  OS=ubuntu, DEPS_FROM=pip, MDTRAJ=mdtraj-release, CONDA_PY=3.8
  OS=ubuntu, DEPS_FROM=pip, MDTRAJ=mdtraj-release, CONDA_PY=3.9
  OS=ubuntu, DEPS_FROM=pip, MDTRAJ=mdtraj-dev, CONDA_PY=3.9
  OS=windows, DEPS_FROM=conda, MDTRAJ=mdtraj-release, CONDA_PY=3.7
  OS=windows, DEPS_FROM=conda, MDTRAJ=mdtraj-release, CONDA_PY=3.9
  ```

The design of this matrix is discussed in #119.


dwhswenson:
In working on this, I noticed that we're doing some ugly stuff in requirements/installation handling. CI is still using a separate `requirements.txt` file (although pip will use the information in `setup.cfg`). Additionally, optional installs are kept in `requirements.txt`-style files in `ci/pip-install`.

As part of this PR, I'll clean that up. The single source of truth will be `setup.cfg`. I'll remove `requirements.txt` and the other files from the repo. I'll add optional requirements as `options.extras_require`.

For now I'm using the names `test` for testing extras (everything required for CI) and `all` to hold all others (everything in `optional_installs.txt`. I didn't include some old coverage tools that we don't use anymore, but which were listed in `ci/pip-install/testing_requirements`.

@sroet : Any thoughts on the names? You can see the change in https://github.com/dwhswenson/contact_map/pull/120/commits/30630304042df5e60c2c990a159296591c75b8e7.

dwhswenson:
Well, that was easy! Once I got the conda install procedure working on Linux, I just had to switch the OS to Windows in the matrix.

Testing requirements need to be pip-installed, because I haven't bothered to put autorelease on conda-forge yet.

This also removed a lot of old files. The `ci` directory is *much* simpler now. There's a single source for requirement versions in setup.cfg.

Also, CI doesn't take nearly as long, and the CI workflows for Linux and Windows are shared (so if there's ever a Windows/conda issue, it is really easy to decouple those things by doing a conda build on Linux and/or a pip build in Windows.)

This is ready for review. It can be merged as soon as @sroet has reviewed it.

ci/pip-pre-install.sh
```diff
@@ -3,4 +3,4 @@
 pip install --upgrade pip
 pip install cython  # may be required for numpy override?
 pip install --upgrade --force-reinstall numpy  # override Travis numpy
-pip install -r requirements.txt
+#pip install -r requirements.txt
```
sroet, line 6: leftover old code?
```suggestion
```

ci/extract_requirements.py
```diff
@@ -0,0 +1,42 @@
...
+
+import argparse
+import os.path
+
+def make_parser():
+    default_setup = os.path.join(os.path.dirname(__file__), "..",
+                                 "setup.cfg")
+    parser = argparse.ArgumentParser()
+    parser.add_argument('setup_cfg', nargs='?', default=default_setup)
+    # parser.add_argument('--extra', default=None)
```
sroet, line 20: leftover debug statements?
```suggestion
```

sroet (approved):
LGTM. Found two commented out lines of code, that should probably be removed. Feel free to ignore

sroet:
> @sroet : Any thoughts on the names? You can see the change in 3063030.

Your names seem great. I could not think of better names for now

ci/extract_requirements.py
dwhswenson: Actually, stolen code from OPS 😆

PR #121, opened 2021-07-29
also drop appveyor badge

sroet:
after #120 we should also drop the corresponding appveyor status badge from the readme


dwhswenson (approved):
Yep! Merging.

PR #123, opened 2022-01-03
update copyright to 2022

sroet:
Happy new year!

I `grep`-ed for `2020` and `2021` and this was the only file that came up.


dwhswenson (approved):
LGTM -- merging!

PR #126, opened 2022-09-16
fix matplotlib 3.6 'Unable to determine Axes to steal space for Colorbar

sroet:
With the release of matplotlib `3.6.0` we triggered an error because we relied on [deprecated behavior](https://github.com/matplotlib/matplotlib/blob/d04c8de4a89ab756a137ef146e8542e81041909f/lib/matplotlib/figure.py#L1182-L1191)
(We did not trigger the deprecation warning as our `mappable` does not have the attribute `axes`)

This reinstates the old behavior (grabbing the current `axes` via `plt.gca()`)

### DIFF
--- a/contact_map/plot_utils.py
+++ b/contact_map/plot_utils.py
@@ -22,6 +22,7 @@ _SEQUENTIAL = [
     'afmhot', 'gist_heat', 'copper'
 ]
 
+
 def is_cmap_diverging(cmap):
     if cmap in _DIVERGING:
         return True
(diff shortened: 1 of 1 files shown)


sroet:
@dwhswenson this one is a one-liner (quick to review) and absolutely required to use any of our plotting with matplotlib `3.6.0`, so I hope you will have some time to look at it

sroet:
I re-enabled unit-tests, let's see if closing and reopening triggers CI

dwhswenson (approved):
LGTM. Thanks!

Nightlies had stopped on this, too? I recently went through a ton of repos and restarted nightly builds. Such a hassle; I may end up writing a local cron job that checks that nightlies are still running on my still-maintained repos.

PR #128, opened 2023-08-08
Update RTD config

dwhswenson:
Apparently docs were not building correctly, and furthermore, we had some setting to use system packages, which will no longer be allowed.

Let's get that fixed.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,16 +1,14 @@
 version: 2
 
-#conda:
-    #environment: docs/environment.yml
+build:
+  os: ubuntu-22.04
+  tools:
+    python: 3.11
 
 python:
     install:
       - requirements: docs/rtd_requirements.txt
-      #- requirements: requirements.txt
       - requirements: docs/requirements.txt
       - method: setuptools
         path: .
-      #- method: pip
-        #path: .
-    #system_packages: false
-
+    system_packages: false


.readthedocs.yml
```diff
@@ -1,16 +1,14 @@
...
       - requirements: docs/rtd_requirements.txt
-      #- requirements: requirements.txt
       - requirements: docs/requirements.txt
       - method: setuptools
         path: .
-      #- method: pip
-        #path: .
-    #system_packages: false
-
+    system_packages: false
```
sroet, line 14: isn't this the default anyway (and thus can be left out)? Assuming I read [this](https://docs.readthedocs.io/en/stable/config-file/v2.html#python-system-packages) correctly?
```suggestion
```

sroet (changes requested):
1 unnecessary line seems to be added, otherwise LGTM

.readthedocs.yml
dwhswenson: For now, I want it there to ensure we're overriding anything set once-upon-a-time in the web interface. (The existence of the YAML file should override web settings, but this is just to be extra careful.) The specific event that triggered this PR was an email from RTD with the following paragraph:

> **We are removing the "use system packages" feature on August 29th.** Make sure you are installing all the required dependecies to build your project's documentation using a `requirements.txt` file and specifying it in your `.readthedocs.yaml`.

dwhswenson:
NB: this is still failing in RTD, so it needs some more work. @sroet : I noticed that you aren't listed as a maintainer with RTD; an oversight because I so rarely need to log into the RTD interface. Do you use the same username over there?

.readthedocs.yml
sroet: seems reasonable to me

sroet:
> Do you use the same username over there?

I do

sroet (approved):
LGTM, feel free to merge if this works on RTD

sroet (approved):
Seems like that build passed, feel free to merge!

PR #124, opened 2022-02-15
use new way of checking for no warnings

sroet:
our test suite started warning:
```
contact_map/tests/test_contact_count.py::TestContactCount::test_pixel_warning
  /home/sroet/miniconda3/envs/contact_map/lib/python3.9/site-packages/_pytest/python.py:192: PytestRemovedIn8Warning: Passing None has been deprecated.
  See https://docs.pytest.org/en/latest/how-to/capture-warnings.html#additional-use-cases-of-warnings-in-tests for alternatives in common use cases.
    result = testfunction(**testargs)
```

This updates the "no warning" check to the new recommendation in [that link](https://docs.pytest.org/en/latest/how-to/capture-warnings.html#additional-use-cases-of-warnings-in-tests)

PR #133, opened 2024-01-28
tests: `setup`,`teardown` -> `setup_method`, etc

dwhswenson:
Pytest prefers (and now, apparently requires) the use of `setup_method` instead of `setup`.

Looks like there was also a problem to fix about newer pytest not liking `pytest.warns(None)` as a way to indicate that no warnings were issued.

### DIFF
--- a/contact_map/tests/test_concurrence.py
+++ b/contact_map/tests/test_concurrence.py
@@ -65,7 +65,7 @@ class ContactConcurrenceTester(object):
 
 
 class TestAtomContactConcurrence(ContactConcurrenceTester):
-    def setup(self):
+    def setup_method(self):
         self.concurrence = AtomContactConcurrence(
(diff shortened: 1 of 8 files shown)


dwhswenson:
This is ready for review and comment. Without review, I will merge this after at least 24 hours, no sooner than Mon 29 Jan 17:00 GMT (11:00 my local).

dwhswenson:
Noting here that the `pytest.warns(None)` issue had been previously solved by @sroet in #124, which I hadn't merged. I merged that one, hoping that credit will go where it is due. FWIW, the only difference (character-by-character) in our solutions were a single vs double quote (took @sroet's single) and that I had added a comment (kept my additional comment.)

sroet (approved):
LGTM, feel free to merge

PR #135, opened 2024-07-19
Update Python versions in CI

dwhswenson:
Aside from being overdue and needed, this is largely in response to a CI failure due to a change in mdtraj-dev. If this is just that our Python CI is out of date, then the problem is on us. If we're still failing after this PR, then we'll take it upstream.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -23,9 +23,9 @@ jobs:
     strategy:
       matrix:
         CONDA_PY:
+          - "3.13"
           - "3.12"
           - "3.11"
-          - "3.10"
         OS: ["ubuntu"]
         MDTRAJ: ["mdtraj-release"]
         INSTALL_WITH: ["pip"]
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -23,22 +23,22 @@ jobs:
     strategy:
       matrix:
         CONDA_PY:
-          - 3.9
-          - 3.8
-          - 3.7
+          - "3.12"
+          - "3.11"
+          - "3.10"
         OS: ["ubuntu"]
         MDTRAJ: ["mdtraj-release"]
         INSTALL_WITH: ["pip"]
         include:
-          - CONDA_PY: 3.9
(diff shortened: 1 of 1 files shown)


dwhswenson:
Sent upstream in https://github.com/mdtraj/mdtraj/pull/1913.

dwhswenson:
Been having more failures lately my guess is incompatibility of new `ubuntu-latest` with some old Python versions. This seems to fit, and brings makes us compliant with SPEC0.

This is ready for review and comment. Without review, I will merge this after at least 24 hours, no sooner than Sat 21 Dec 15:00 GMT (09:00 my local).

.github/workflows/unit-tests.yml
```diff
@@ -23,22 +23,22 @@ jobs:
...
             OS: "ubuntu"
             MDTRAJ: "mdtraj-dev"
             INSTALL_WITH: "pip"
-          - CONDA_PY: 3.9
+          - CONDA_PY: "3.12"
             OS: "windows"
             MDTRAJ: "mdtraj-release"
             INSTALL_WITH: "conda"
-          - CONDA_PY: 3.7
+          - CONDA_PY: "3.10"
```
sroet, line 41: Any reason to test outside of the spec0 matrix?

sroet (approved):
1 question, LGTM otherwise

.github/workflows/unit-tests.yml
sroet:
```suggestion
          - CONDA_PY: "3.11"
```

.github/workflows/unit-tests.yml
```diff
@@ -23,22 +23,22 @@ jobs:
...
         MDTRAJ: ["mdtraj-release"]
         INSTALL_WITH: ["pip"]
         include:
-          - CONDA_PY: 3.9
+          - CONDA_PY: "3.12"
             OS: "ubuntu"
             MDTRAJ: "mdtraj-dev"
             INSTALL_WITH: "pip"
-          - CONDA_PY: 3.9
+          - CONDA_PY: "3.12"
```
dwhswenson, line 37:
```suggestion
          - CONDA_PY: "3.13"
```

PR #136, opened 2026-01-24
Wrap json strings in io.StringIO in deserialize

dwhswenson:
Looks like pandas 3.0 has changed things so that it prefers `read_json` to think its input is a filename; we've been assuming it was the contents of that file. Hoping this fixes (had trouble reproducing locally for some reason?)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -71,3 +71,6 @@ docs/_build/
 
 # PyBuilder
 target/
+# pixi environments
+.pixi/*
+!.pixi/config.toml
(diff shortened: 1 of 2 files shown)


dwhswenson:
Tests were failing for the last 2 days, and this seems to fix it. Calling that worthy of review.

This is ready for review and comment. Without review, I will merge this after at least 24 hours, no sooner than Sun 25 Jan 03:00 GMT (Sat 24 Jan 21:00 my local).

